PR #110, opened 2014-10-18
Code box improvements

Ragnar-F:
This shows the buttons select and download in lower right corner of a pre box.

Select button will select the complete code so that the user can copy it into the clipboard. Tested with Chrome, Firefox and IE8+.

Download button will download the complete code into a file with the format "[url file name]-Script.ahk"; including a BOM. Works only in HTML5-able browsers (because of the added download attribute) and IE10+.


Lexikos:
I also made some icons to use in place of the text, but the end result didn't seem like it would be worth the effort to get them working properly. (By "properly" I mean not adding screen noise to every code box, just the one being hovered over.)

Download Select

joedf:
Neat little icons there :+1:

Ragnar-F:
Thanks for merging. Nice idea with the icons.

PR #113, opened 2015-01-15
Fixed misbehavior of index list when single-clicking

Ragnar-F:


Lexikos:
What's this "misbehaviour"?  Why should it need a double-click?

Ragnar-F:
1) Offers the possibility to focus the listbox with the mouse and navigate with the keyboard (E.g. via the arrow keys or the initial character of each list entry), without opening a new page by mistake (which can be really disturbing for a page with an anchor, because it automatically scrolls down and therefore the listbox is no longer visible).

2) For example, if I write "loop" in the editbox and click "Loop (Registry)", "Loop" will open. This is obviously an incorrect behavior.

3) If I would replace "change" with "click" instead of "dblclick", it would eliminate issue #2, but not issue #1. Additionally, when single-clicking the editbox to write something in, it would open the selected listbox entry immediately.

Lexikos:
Okay.

1) That highlights another problem; that the index is anchored to the top of the page. I see no reason it couldn't be placed statically on the screen.

Ragnar-F:
I tried already to make it statically (like the sidebar in the chm), but I failed :( especially in the case making the whole sidebar scrollable, if the height of the browser is lower than the height of the sidebar. If I have time, I'll look into this again.

Lexikos:
Commit 922522a0 fixes the index in place.

It could probably be applied to the TOC as well, but I tend not to use the TOC. "Gluing" the TOC to the top and bottom of the page the way [MSDN](http://msdn.microsoft.com/en-us/library/ie/hh801964.aspx) does it is another idea, but that behaviour sometimes annoys me.

Ragnar-F:
Really nice. Thanks. I think the same about the toc.

PR #116, opened 2015-01-22
Change data files (toc, index) into pure JSON files 

Ragnar-F:
Since the data are almost in JSON format, this would be the next logical step.

Also 	move translatable items to a separate file (so I don't have to translate the js code).

The only disadvantage I can see is that the file size is increased (ca. 8 %), because the key names also must be enclosed with quotation marks.


Lexikos:
Why? What's the *advantage*?

Ragnar-F:
In addition to the general benefits found on http://en.wikipedia.org/wiki/JSON (in short: language-independent data format, human-readable text, standardized format), also personal benefits (respectively for translators), because in this way there is a clear separation between code and data. As a result, I can better organize my files and use a json format support (okapi plugin in this case) in the translation program, instead of creating regex "hacks", which is cumbersome, time consuming and only partially working (due to the limited capabilities of the translation program, as you already have noticed - I don't know better ones).

Lexikos:
> the general benefits found on http://en.wikipedia.org/wiki/JSON (in short: language-independent data format, human-readable text, standardized format)

That equates to "no advantage". This data is for consumption by the documentation. JSON might be considered "language-independent", but that only means people have written parsers for it in multiple languages. It was already human-readable and in a standardized format.

> I can better organize my files and use a json format support (okapi plugin in this case) in the translation program

That sounds more useful. If it helps you and other translators in a meaningful way, I'm more inclined to merge it.

> instead of creating regex "hacks", which is cumbersome, time consuming and only partially working

Couldn't you just load the ToC/index and produce JSON files for your translation program? That would work just the same as having the original data in JSON format. You don't even need to parse the data; just do `JSON.stringify(toc)`. For arguments sake, I wrote [a simple HTA which does the job](https://gist.github.com/Lexikos/5f8e1402943642ff98c7). (However, I'm not really proposing that you use it.)

To reverse the increase in file size, could the data be changed to use arrays? I'm guessing that jqTree requires that specific format for the ToC, but for the index there doesn't seem to be any reason to use `{"t":"Topic","v":"URL"}` over `["Topic","URL"]`.

Actually, all of the URLs begin with "docs/". This prefix could be removed.

Lexikos:
Then again, I see no reason that CreateFiles4Help.ahk couldn't strip out the unneeded quote marks when it inserts the data into content.js.

Ragnar-F:
Never mind, I discard the idea of using JSON. Instead, I rename the data containing files to data_*.js, which would helping me, too. I guess this small change shouldn't make any problems.

Lexikos:
For the record: If there's an actual (practical) benefit to offset the increase in file size (or if the generated script is not any larger), I'm not against using JSON.

PR #117, opened 2015-01-25
Fix umlaut display issue in chm file

Ragnar-F:


Lexikos:
I was going to say "what issue?" because I'd already [fixed it](http://ahkscript.org/docs/AHKL_ChangeLog.htm), but then realized that I hadn't committed the fix.

That text comes from the git commit history, which is not HTML. Each time I release an update, my build script takes the git history, does some rudimentary regexreplacing and shoves it into a temporary file for me to edit. When I'm done editing (and it's done building), it shoves the edited HTML into the changelog. This time I think the encoding error was caused by my build script.

PR #118, opened 2015-01-25
Force reading files as UTF-8

Ragnar-F:
... to prevent showing non-english characters as gibberish inside the navbar of chm or online help.


Lexikos:
But there aren't any non-ASCII characters in those files...

If there were, one would have to be sure to encode the files as UTF-8. If you used UTF-8 with BOM, FileEncoding would be redundant.

Issue #122, opened 2015-08-06
Ahk2Exe - Mention of the Error codes


joedf:
Merged into AutoHotkey/Ahk2Exe since v1.1.22.03.
Maybe a mention of the error codes ([ErrorCodes.md](https://github.com/fincs/Ahk2Exe/blob/master/ErrorCodes.md)) would be nice.  

On this page, http://ahkscript.org/docs/Scripts.htm#ahk2exe

PR #123, opened 2015-08-11
Ahk2Exe - Mention of the Error codes #122

joedf:
see https://github.com/Lexikos/AutoHotkey_L-Docs/issues/122

### DIFF
--- a/docs/Scripts.htm
+++ b/docs/Scripts.htm
@@ -149,6 +149,7 @@ Usage:
   <li>The method above can also be used to change existing icons or add new ones to all compiled scripts.</li>
   <li>The built-in variable A_IsCompiled contains 1 if the script is running in compiled form. Otherwise, it is blank.</li>
   <li>When parameters are passed to Ahk2Exe, a message indicating the success or failure of the compiling process is written to stdout. Although the message will not appear at the command prompt, it can be &quot;caught&quot; by means such as redirecting output to a file. <span class="ver">[v1.0.43+]</span></li>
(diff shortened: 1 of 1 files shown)


Lexikos:
Errors which I've fixed: occured, ErrorCode.md

The text you've added is open to interpretation, like "In the case of failure, Ahk2Exe has exit codes." Actually, Ahk2Exe always has exit codes - but when are they *used*? If there is a failure, is the exit code *always* one of those values? If the exit code is one of those values, does it necessarily mean the corresponding error occurred and compilation failed? What happens if compilation succeeds?

Typically the exit code would be used by another script, and the author wants to know most of all how to determine whether compilation succeeded or failed. So it's better to say what the exit code will be in *both* cases. Unfortunately, something like "a non-zero exit code indicates success" would be incorrect, since Ahk2Exe still returns zero without compiling in some cases, such as if AutoHotkeySC.bin is missing.

I think it's clearer for "the exit code" to mean the single number which the process returns on exiting once, and "error codes" to mean any or all of the possible values.

joedf:
Agreed. I have added return `0x0` for `success`. All `exitapp` commands are covered except for `guiexit`. I have tested with this [commit](https://github.com/fincs/Ahk2Exe/issues/33). When missing `AutoHotkeySC.bin`, it should return an error. eg. if the compilation is cancelled: it will return `0x2`.

joedf:
Much simpler?
> Additionally, Ahk2Exe has exit codes. These error codes can be found at GitHub (ErrorCodes.md).

PR #126, opened 2015-11-23
Adding prefix "My" to examples

T-Rock:
Make Differentiation more obvious between using a word literally or as an example by adding prefix "My".


joedf:
Interesting idea to reduce possible confusion...

PR #128, opened 2016-01-22
Switching to HTML5

Ragnar-F:
Changed doctype to officially support HTML5 (previously it was HTML 4.01), mainly to get rid of the long doctype line and some barely visible position issues in the navbar since it originally based on html5.

And validated the documents via W3C validator to make them fit for html5. The most significant change can be found in Progress.htm, where I changed the color table structure to match with the other ones. Other changes are converting "<a name=...>" anchors to "id=..." and removing unused ones, and replacing obsolete attributes like "width=25%" with "style='width:25%'" etc.

I fast-tested the results and saw no huge issues. All in all we should save some bytes.


Lexikos:
Thanks for this. However...

What do you mean by *unused* anchors? Anchors are used from more places than just the documentation, like the forum. If you remove or rename anchors, you will break links. There's no good reason to do that.

Ragnar-F:
I had situations where 2 anchors for 1 target exist. Since I'm not allowed to put multiple id attributes in one tag (because only the first one is in effect), I was "forced" to choose one and discard the other one.

But nevermind. I found a workaround. You can see it if you search for `<span id="..."></span>`.

Ragnar-F:
Beside of the merge conflict, what's wrong here?

Lexikos:
Hmm, thought I had replied.  I had been holding off on pushing the updates so that I could use them to test a GitHub web hook, but it's not ready yet.  The changes were pushed as a result of releasing v1.1.23.04.

Thanks.

PR #129, opened 2016-01-22
Add separate InStr page

### LINKED ISSUE #43: Own pages for builtin functions - InStr()
https://trello.com/card/own-page-for-each-built-in-function-function-group/4ff142712c442e040466599d/24

--------

Own page for InStr()

Taken from: http://www.autohotkey.com/community/viewtopic.php?p=502134#p502134

Ragnar-F:
This should resolves #43. I was thinking of including this page into IfInString, but then I noticed that it would also fit into StringGetPos, therefore I took the happy medium and created a new site (based on the creations of my forerunners).


Lexikos:
Good work. There were a couple of "copy paste paste" errors in the StartingPos parameter which I have fixed.

PR #130, opened 2016-02-11
Add SubStr page

### LINKED ISSUE #42: Own pages for builtin functions - SubStr()
https://trello.com/card/own-page-for-each-built-in-function-function-group/4ff142712c442e040466599d/24

--------

Own page for SubStr()

Template: Taken from: http://www.autohotkey.com/community/viewtopic.php?p=502134#p502134

Ragnar-F:
This should resolves #42. I decided against it merging SubStr with StringMid, because I think it's better to ignore StringMid as far as possible, since it will be removed in v2. In other words, it's not worth it. Additionally in this way it's clearer in my opinion.


Lexikos:
Thanks.

PR #131, opened 2016-03-11
Clarifying "enable" and some minor layout changes.

Micromegas2016:
1. Using "enable" per [Discussion in forum](https://autohotkey.com/boards/viewtopic.php?f=14&t=14772). 
2. Moving blank to end of list, since it is not a sub-entry for `Period`. 
3. Replacing "execute" with "jump" to be consistent with terminology used in [Label section](https://autohotkey.com/docs/commands/SetTimer.htm#Label). (It's also more precise, since execution can take considerable time. This choice of words should thus reduce the need for comments such as "If a timer is disabled while its subroutine is currently running, that subroutine will continue until it completes.".) 
4. Adding strong text at beginning of subsections, as has been done for "_Timer Precision_" already 
5. Deleting "unless it is still running from the last time" from "Period" subsection since it is not about that parameter, and it's already treated in "Reliability" section. 
6. Moving `Period` range condition to `Period` definition.


Lexikos:
Superseded by cb9e93f7.

PR #135, opened 2016-06-16
Fix SubStr's related links for StringLeft/Right and StringTrimLeft/Right

G33kDude:


joedf:
? ...  :0
https://github.com/Lexikos/AutoHotkey_L-Docs/blob/master/docs/commands/StringLeft.htm
https://github.com/Lexikos/AutoHotkey_L-Docs/blob/master/docs/commands/StringTrimLeft.htm

G33kDude:
`SubStr.htm` is already in the commands folder. With the prefix in the html, it tries to go to `docs/commands/commands/StringLeft.htm`, which errors.

joedf:
Cool :+1:
I missed that haha

PR #139, opened 2016-07-19
Fix v2 code examples

Ragnar-F:
... and also checked if they work. There are a some examples which don't work, even though the syntax is correct (I think).  Please take a look at them, maybe you can find the problem:

**docs\commands\RegisterCallback.htm**

    callback := RegisterCallback("TheFunc", "F", 1)  ; Parameter count must be specified.
    TheFunc("TheFunc was called directly.")          ; Call TheFunc directly.
    DllCall(callback, float, 10.5, int64, 42)        ; Call TheFunc via callback.
    TheFunc(params*) {
        if IsObject(params)
            MsgBox % params[1]
        else
            MsgBox % NumGet(params+0, "float") ", " NumGet(params+A_PtrSize, "int64")
    }

**docs\commands\Sort.htm**

    MyVar := "1,2,3,4"
    Sort, MyVar, %MyVar%, F ReverseDirection D,  ; Reverses the list so that it contains 4,3,2,1
    ReverseDirection(a1, a2, offset)
    {
        return offset  ; Offset is positive if a2 came after a1 in the original list; negative otherwise.
    }

**docs\commands\Until.htm**

    Loop, Read, %A_ScriptFullPath%
        lines .= A_LoopReadLine . "`n"
    Until A_Index=5  ; Read the first five lines.
    MsgBox % lines

**docs\commands\VarSetCapacity.htm**

    ; Calculate required buffer space for a string.
    max_chars := 500
    max_bytes := max_chars * 2
    
    Loop 2
    {
        ; Allocate space for use with DllCall.
        VarSetCapacity(buf, max_bytes)
    
        if A_Index = 1
            ; Alter the variable indirectly via DllCall.
            DllCall("wsprintf", "ptr", &buf, "str", "0x%08x", "uint", 4919)
        else
            ; Use "str" to update the length automatically:
            DllCall("wsprintf", "str", buf, "str", "0x%08x", "uint", 4919)
    
        ; Concatenate a string to demonstrate why the length needs to be updated:
        wrong_str := buf . "<end>"
        wrong_len := StrLen(buf)
    
        ; Update the variable's length.
        VarSetCapacity(buf, -1)
    
        right_str := buf . "<end>"
        right_len := StrLen(buf)
    
        MsgBox,
        (
        Before updating
          String: %wrong_str%
          Length: %wrong_len%
    
        After updating
          String: %right_str%
          Length: %right_len%
        )
    }

**docs\misc\ImageHandles.htm**

    ; Show a menu of the first n files matching a pattern, and their icons.
    pattern := "%A_ScriptDir%\*"
    n := 15
    
    ; Allocate memory for a SHFILEINFOW struct.
    VarSetCapacity(fileinfo, fisize := A_PtrSize + 688)
    
    Loop, Files, %pattern%, FD
    {
        ; Add a menu item for each file.
        Menu F, Add, %A_LoopFileName%, donothing
        
        ; Get the file's icon.
        if DllCall("shell32\SHGetFileInfoW", "wstr", A_LoopFileFullPath
            , "uint", 0, "ptr", &fileinfo, "uint", fisize, "uint", 0x100)
        {
            hicon := NumGet(fileinfo, 0, "ptr")
            ; Set the menu item's icon.
            Menu F, Icon, %A_Index%&, HICON:%hicon%
            ; Because we used ":" and not ":*", the icon will be automatically
            ; freed when the program exits or if the menu or item is deleted.
        }
    }
    until A_Index = n
    Menu F, Show
    donothing:
    return


Lexikos:
Thanks for doing this.

RegisterCallback: Seems to work fine for me.

Sort, Until, ImageHandles: This is due to bugs in v2.0-a075.  I'll fix them.

VarSetCapacity: The syntax error is clearly pointed out by an error dialog. If I escape the percent sign properly the example seems to work fine. See 9d0b9fe3.

For the FAQ (which I think does *not* fall under "Fixed code examples"), I've avoided reviewing them with v2 in its current state. The percent sign FAQ which you updated probably needs more drastic change; I think it's more logical to say "variable are enclosed in percent signs when written within text/a string" than "variables are always enclosed in percent signs except when A, B, C and D", where the exceptions cover most of the language... Anyway, that's a bit off-topic.

Ragnar-F:
Just to clarify, I'm using v2.0-a075 32-Bit on Windows 10 64-Bit.

The RegisterCallback example shows the two message boxes as intended, but then the following error message occurred:

	Error:  Parameter list too large, or call requires CDecl.
	
	Specifically: 8
	
		Line#
		002: TheFunc("TheFunc was called directly.")
		003: DllCall(callback, float, 10.5, int64, 42)
		004: {
		005: if IsObject(params)
		006: MsgBox,params[1]
		007: Else
		008: MsgBox,NumGet(params+0, "float") ", " NumGet(params+A_PtrSize, "int64")
	--->	009: }
		010: Exit
		011: Exit
		011: Exit
	
	The thread has exited.

VarSetCapacity: I don't know why I missed that. Really easy to fix. :(

Regarding to the FAQ, you're right. I just wanted to preserve the sentence structure as far as possible to prevent too much irrelevant changes based on this commit.  Unfortunately The result is far from optimal.

Lexikos:
Ah, that's because the 32-bit callback only accepts 4 bytes of parameters, and we gave it 12.

It's not an issue on x64 because the x64 calling convention is like cdecl; the callback accepts any number of parameters, and the caller is responsible for cleaning up afterwards. That also means there's nothing for DllCall to check, so no exception if the number of parameters was incorrect (unless the function itself throws an SEH exception).

The example has been broken since v1, but on v2 DllCall throws an exception instead of setting ErrorLevel. I don't know what I was thinking with specifying the parameter count as 1. It should be the size of the parameter list. See 40a9e8a6 (v1 branch).

I already updated the FAQ.  See c9f668bd.

Ragnar-F:
Unterstood. Thanks for the explanation.

PR #146, opened 2016-08-08
Add GetKeyName/VK/SC page

Ragnar-F:
Just an idea for v2:

GetKeyState, GetKeyName, GetKeyVK, GetKeySC should be renamed to KeyGetState, KeyGetName, KeyGetVK, KeyGetSC, to be consistent with command names like WinGetX.

Then we could merge the docs from this commands into KeyGet.htm, similar to WinGet.htm.


Lexikos:
There's no need to have index entries for both GetKeyState and GetKeyState(), pointing to the same place. I've removed one.

**Re your idea:**

If it makes sense to group related commands/functions on a single page, we can do that regardless of what they are named. Replacing "GetKey" with "KeyGet" doesn't make any real difference.

WinGet.htm is a poor example to follow. Why are some WinGet functions on that page and several others not? There are two reasons:

  1. They were grouped there because they were sub-commands, and the other commands were not. Chris probably had his reasons for making them sub-commands in the first place (used more rarely?), but IMO, the division isn't very rational.

  2. When I updated the v2 docs, I intentionally limited my changes to what's important (content, accuracy), to avoid burning out what little motivation I had.

I don't want to go into great length here because I'm closing this pull request (and it's a bit off-topic), but I'm going to dump some ideas here anyway:

I don't necessarily agree with grouping commands together on a single page. In some cases it would be better to have a page for each function (for quick reference) and a page for the "category" (as in some wikis) or related concept, which lists all related functions. For example, a page about "window" concepts would be a good central place to explain some concepts that aren't really explained currently (like defining "windows" and "controls/child windows"), and could list all window-related commands, describing "properties of a window" alongside the corresponding WinGet and/or WinSet commands.
For some functions, like GetKeyName/VK/SC, they're so similar that separate pages would just be a waste of space.

Ragnar-F:
So basically a document for each TOC entry with subitems, like "Registry Management" or "Window Management".

In the old wiki, I categorized the ahk docs in conjunction with the pattern `[[Category:Name]]`, which was very convenient. It supports multiple categorization, so for example I could define the categories "Command" (for each command/function, it produced basically the command list), "Function" and "String" for RegExMatch.

So one question is, where the category docs should be. In the commands folder or in a separate folder?

Lexikos:
There are already general topics in / and /misc/, although I suppose that even the current content could be organised better.  I think that the file system isn't particularly important; the structure within the documentation is.

PR #147, opened 2016-08-09
Add IsByRef / IsFunc / IsLabel / IsObject page

Ragnar-F:


Lexikos:
I notice that you are tending to use the *OutputVar* concept for return values. I don't like it (but I'm not motivated to change it). I touched on it briefly in the [Commands vs Functions](https://autohotkey.com/boards/viewtopic.php?f=37&t=19457) topic:

>   "OutputVar: The name of the variable in which to store xyz." When the syntax is shown together with `StrLen, OutputVar, String` and `OutputVar := StrLen(String)`, this description works with both, but **it misrepresents the capabilities of expressions** (implies that you have to use a variable).

Aside from that, the reader has to pass over several words (describing a common piece of syntax that the reader should already know about) to get to the part that's actually relevant to this function. It might even make learning the "return value" concept more difficult.

The convention with functions (even used with some of the documentation you have removed from the Functions page) is to describe the *return value*, as in "Function(a, b): Returns xyz.", or under a "Return Value" heading (NumGet, NumPut, FileOpen, ComObjCreate, ComObjType, etc.).

The terminology used for the IsXxx() functions is inconsistent.  We have "Determines whether", "Indicates whether", "Checks whether" and "Determines if".  I'm not sure what's best.  What does the function do after it determines/checks whatever?  It "Returns a (boolean) value indicating whether ..." but that's a bit verbose.  Some of the original documentation used "Returns x if", which is clear and succinct.

"IsFunc: Indicates whether a function may be called dynamically." is incorrect.

  - *All* functions can be called dynamically if the usage is correct, and IsFunc doesn't check your usage.
  - IsFunc returning false does not mean the value cannot be "called dynamically".
  - IsFunc tells you whether the value is a function, where "function" in this case really means a function (built-in or created by a function declaration in the script), not some other callable object. `if Func(FunctionName)` and `if IsFunc(FunctionName)` have the same result when FunctionName is a string. (There probably isn't much reason to keep it in v2.)

Ragnar-F:
I agree with you and will implement your suggestions. The descriptions for IsXxx() functions were not my idea. I've just used the ones from the command list and therefore assumed that they were already more or less thought out.

Lexikos:
I'm not pointing fingers, just hadn't noticed the issues before now. I don't use the command list.

PR #149, opened 2016-08-12
Move return value to own section

Ragnar-F:


Lexikos:
I think that the syntax/usage block at the top should show the common usage, i.e. with an assignment to catch the return value.  It's just the "OutputVar" and "The name of the variable in which to store" that I don't like.

A search of `class="Syntax">\w+ :=` will get plenty of examples.  Sometimes there's a short entry in the parameter list (just describing the value, not the syntax), otherwise the variable name itself serves to show at a glance what the function returns.

There's some inconsistency, but addressing it all while the future of v2 is up in the air would probably not be worthwhile.

Ragnar-F:
Fixed.

Lexikos:
Thanks.

> Edit: Btw, `a attribute` should be `an attribute`.

Actually, "an attribute string of the file" sounded a bit strange (as if there are multiple possible values and it returns an arbitrary one), so I changed it to how it was originally, "**the** attribute string".

PR #152, opened 2016-09-05
update SetDefaultMouseSpeed

mmikeww:
'Click' has no speed option to override this setting, so correct the first line of the help page


Lexikos:
Your edit is not more correct. Click does use the speed set by SetDefaultMouseSpeed, and your new text does not reflect that. If you *can't* specify the speed when calling Click (because it has no option), that means it will **always** be unspecified and so the value set by SetDefaultMouseSpeed will be used as documented.

mmikeww:
ok

PR #154, opened 2016-10-20
Uncommented missing links

Ragnar-F:


Lexikos:
v2.0-a076 does include a help file, but there's a larger issue here: the entire Section 2 (Hotkeys & Hotstrings) is still commented out.

See a4b8a6a5.

PR #155, opened 2016-10-28
update if-expression example

mmikeww:
this line `if (A_Index > 100 or Done)` doesn't really make sense and probably not intended to be like that

further down there is an `or` example comparing two strings so probably this can be removed so theres no ambiguity


Lexikos:
None if the if-expression examples make sense out of context, as they are; that one no more or less. Within the context of a loop where a variable `Done` signals that it is finished, it makes perfect sense. There is nothing "ambiguous" here either.

I don't see any problem with your change.  Maybe it will help someone.

PR #157, opened 2016-11-16
update StrLower docs for InputVar param change

mmikeww:
the parameter is now a string, not the 'name of a variable'


Lexikos:
There are likely to be many broken examples throughout the v2 documentation. I mostly only updated examples where I had some other reason to be editing. This is partly due to my "energy budget" and partly because the syntax for v2 hasn't been finalised.

PR #158, opened 2016-11-16
broken StrSplit() link

mmikeww:
the StrLen() link is broken too, v2-changes say the command was removed, but was the func removed?

### DIFF
--- a/docs/Functions.htm
+++ b/docs/Functions.htm
@@ -310,7 +310,7 @@ path-to-the-currently-running-AutoHotkey.exe\Lib\  <em>; Standard library.</em><
     <td>Retrieves the count of how many characters are in a string.</td>
   </tr>
   <tr id="StrSplit">
-    <td><a href="commands/StringSplit.htm">StrSplit</a></td>
+    <td><a href="commands/StrSplit.htm">StrSplit</a></td>
     <td>Separates a string into an array of substrings using the specified delimiters.</td>
   </tr>
   <tr id="WinActive">


Lexikos:
The StringLen.htm page was removed long ago, and until recently the Functions page told everything one needed to know about StrLen().  Now that part of the Functions page is just a redundant index, with short descriptions.

It could have its own page (named StrLen.htm), or we could just remove the broken link. The v2 index points at Functions.htm#StrLen, not StringLen.htm.

mmikeww:
> The v2 index points at Functions.htm#StrLen, not StringLen.htm.

I'm not seeing that. It looks like its pointing to a broken `StringLen.htm`:

https://lexikos.github.io/v2/docs/Functions.htm#BuiltIn

Lexikos:
Yes, that's the broken link that I suggested we could remove...

It's not just the index, but also every other page that links to StrLen.  Only that one link is broken, because it was recently merged from v1.

mmikeww:
I don't know if you use command line git, but you could do `git grep StrLen` or `git grep StringLen` to find other instances

Lexikos:
I thought maybe you were insinuating that there are other instances of StringLen.  There aren't, as I said.  I knew because I had already done a search.

Maybe you misunderstood my comment; the index (data_index.js) and every other page that links to StrLen uses the correct URL, Functions.htm#StrLen.  There is only one link to StringLen.htm; one that snuck in via content merged from v1.

Feel free to fix it in any way you see fit.

mmikeww:
Gotcha, I did misunderstand

PR #161, opened 2016-12-12
Clarify Loop,Count param

mmikeww:
The phrase "can be an expression" was used throughout the v1 docs, and implied "but might not be an expression". That implication is not true for the Loop command in v2. This code works fine on v1 but not on v2, and hopefully this change in the docs clarifies that a bit:

```autohotkey
var := 2
Loop, %var%
   MsgBox, %A_Index%
```


Lexikos:
If the extra explanation about Count being an expression is necessary, that calls into question whether it should be an expression in the first place.

It doesn't take an expression in v1, and couldn't due to ambiguity.  Chris' notes about v2 suggested "some parameter reordering and disambiguation".  When I did this, it seemed appropriate to let it accept an expression.

Since then, commands were changed to not accept expressions by default.  Loop Count wasn't changed because unlike other commands, it couldn't be used with function syntax.  Now it can, in v2.0-a075, although I consider it an experimental feature.  I can think of reasons to keep Count as an expression, but none that would set Loop apart from every other command.  It is inconsistent.

Putting aside whether Loop will continue to accept an expression;

You seem to think that "can be an expression" implies "or a variable enclosed in percent signs".  It doesn't.  I do not think it necessarily implies "or not an expression" either.  It could mean that some *other* parameters which are numbers can't be expressions, or that a simple number is not necessarily an expression.

"How many times, which can be an expression" indicates that (at least sometimes) "How many times" is itself the expression whereas "How many times, or an expression" sounds like it's presenting two mutually-exclusive possibilities.  If it's not "How many times", what is the expression?

mmikeww:
> If the extra explanation about Count being an expression is necessary, that calls into question whether it should be an expression in the first place.
> ...
> Since then, commands were changed to not accept expressions by default. Loop Count wasn't changed because unlike other commands, it couldn't be used with function syntax. Now it can, in v2.0-a075, although I consider it an experimental feature. I can think of reasons to keep Count as an expression, but none that would set Loop apart from every other command. It is inconsistent.

Yes, I ran into this issue, because I know other Loop commands such as `Loop, Parse` no longer use "the name of a variable", meaning that the syntax is now `Loop, Parse, %var%` instead of without the percents as in v1. So I was surprised to see that regular Loop was different. I agree its inconsistent from my brief knowledge. You'd have a better idea about that.

Then again, I guess its only inconsistent with other v2 `Loop` commands. Because other loop constructs such as `while` and `for` both expect expressions. Upon more thought, I tend to think of a regular `Loop` more in line with those other two constructs. Whereas Loop,Parse and Files and the like I think of as more specialized instances. And I don't really care whether or not the Loop commands can be called with function syntax or not. I don't think of them as commands, but rather control flow constructs

If you changed it to not be an expression, but instead literal text value, then I assume it would be like all other commands in that when called using function syntax, that parameter would then be an expression as usual. 

> You seem to think that "can be an expression" implies "or a variable enclosed in percent signs". It doesn't. I do not think it necessarily implies "or not an expression" either. 

But in v1, every time that "can be an expression" phrase is used, a "variable enclosed in percents" was also acceptable as a parameter input as far as I've encountered. Maybe I've missed something

> It could mean that some other parameters which are numbers can't be expressions, or that a simple number is not necessarily an expression.
> "How many times, which can be an expression" indicates that (at least sometimes) "How many times" is itself the expression whereas "How many times, or an expression" sounds like it's presenting two mutually-exclusive possibilities. If it's not "How many times", what is the expression?

Yeah another thing I noticed was that every time "can be an expression" was used in the v1 docs, it seemed to relate to parameters where numbers were expected, but I never saw it when other values were expected.

I guess technically, a pure integer number itself _IS_ an expression, yes? In which case, my re-wording is redundant.

Lexikos:
> Yeah another thing I noticed was that every time "can be an expression" was used in the v1 docs, it seemed to relate to parameters where numbers were expected, but I never saw it when other values were expected.

It has to do with the cost of retaining backward compatibility. It would have been possible to allow expressions with SetTimer, for instance, but there would have to be exceptions for the literal values `On` and `Off`.  In the source code they are referred to as "numeric params", not "params which can be expressions".  For commands which expect arbitrary strings, it wouldn't have been practical to allow expressions because any valid string could appear to be an expression.

> But in v1

This isn't v1, the documentation hasn't been written with v1 in mind, and v2 doesn't hold to the same patterns as v1.  What you're talking about isn't the meaning of the words, but meaning that you (perhaps erroneously) associate with the phrase based on past observations of a different version of the program.

> Count: a number indicating how many times to perform the loop, or any other expression

That's possibly worse. If `getLoopCount()` *is* "a number indicating [etc.]", then "any **other** expression" would have to be an expression which is not a number. The phrasing "as an expression" might be better.

It would be valid to say "can be an expression and nothing else", so I say that "can be an expression" does not imply one way or the other (that it can or cannot be something else).

mmikeww:
> This isn't v1, the documentation hasn't been written with v1 in mind, and v2 doesn't hold to the same patterns as v1. What you're talking about isn't the meaning of the words, but meaning that you (perhaps erroneously) associate with the phrase based on past observations of a different version of the program.

Yes exactly. Maybe its erroneous to assume that others would make the same mistake as I did. But why not clarify it just in case, when it costs so little? Verbosity in documentation is usually preferred over brevity, if for no other reason than to reduce support/forum requests for help. We are writing paragraphs on a trivial matter, unless of course you plan on changing the param

Lexikos:
So is your initial wording change sufficient in your view?

mmikeww:
Sufficient, yes I think so, but probably not ideal. I'm not exactly sure what's best. Your "as an expression" could work too.

I don't think its too big of a deal right now. And if you change the parameter to a non-expr in the future, then the docs can be changed

Lexikos:
Superseded by commit 618f431.  Almost everything's an expression, so we don't say "can be an expression".

PR #164, opened 2016-12-28
add `is` to the v2 index page

mmikeww:


Lexikos:
It's not a command or function any more than `=` or `+`.

Ideally there would be a structured language overview including all functions (some formerly known as commands) and expression operators, but until then, the listing for `is` in the expression operators table and help index is enough.

Actually, the other operators should probably be added to the help index.  `is` already has an entry because it has its own page.

mmikeww:
Ok. I don't think I've ever opened the local .chm, if thats what you're referring to as the 'help index'.  I always use the online docs. I didn't check the expr operators table, I should have. I see it there.

Lexikos:
That's another thing that needs to be fixed; the online help index (and other scripted functionality) relies on a script having combined all of the js files into one, so doesn't work on GitHub.  There should be a fallback, such as a version of content.js which just loads the other files.

The help index works in the local HTML files if you've run `docs/static/source/CreateFiles4Help.ahk`, but it has to be run each time one of the js files changes.

Lexikos:
> There should be a fallback, such as a version of content.js which just loads the other files.

There is, as of 00699f7. The help index and table of contents now work on GitHub.

Lexikos:
Commit acd14ae adds expression operators to the help index. They will not be added to the command/function list.

PR #165, opened 2017-02-07
Changed the A_OSVersion win10 example

CapnOdin:
from 6.4.9841 to 10.0.14393


Lexikos:
Your proposed text is misleading and incorrect. 10.0.14393 is Windows 10, but Windows 10 is not 10.0.14393. Windows 10 is many different version numbers.

10.0.14393 is also known as version 1607.

CapnOdin:
Would you prefer 10.x.x ?

Lexikos:
> I got a little confused by "TP" I think it means currently but I am not sure.

If you search for the version number (6.4.9841) you should find that TP is short for Technical Preview. It is a specific Windows version. 6.4.9841, obviously.

> Would you prefer 10.x.x ?

No. The reader must realise that x is a placeholder, which is already the case for "major.minor.build", and it still doesn't show what parts of the number may change, or that every part could potentially change.

"For example, `10.0.14393` is Windows 10 build 14393, also known as version 1607."

CapnOdin:
I feel that 6.4.9841 needs to be changed, since one of my scripts misbehave because of the change from 6 to 10. I used it to handle the difference in TrayTips.

I don't know how to properly phrase the change, so I will close the pull request.

Lexikos:
I replaced the example with the last line of my previous comment.

See 946aec8f.

PR #175, opened 2017-03-30
v2: Integrate new GUI API

Ragnar-F:
Some paragraphs or sentences are commented out (`<!-- ... -->`) because I don't know exactly whether they're still relevant for the new API or not. Maybe someone with more expertise should check it. Are the A_Gui* built in vars still functional somehow?

Besides of this, I'm not done yet. Following tasks are still to do:

- [x] Revise commands/TreeView.htm
- [x] Revise commands/ListView.htm
- [x] Revise script showcases


Lexikos:
Thanks. I wasn't looking forward to doing the documentation for this.

No, there are no A_Gui variables. Everything is passed via parameters to the event handler functions.

A_DefaultGui, A_DefaultListView and A_DefaultTreeView are also being removed, of course.

I'll probably wait until I'm finished modifying the Gui API (for a time, at least) and then merge this and review the documentation.

Lexikos:
I think +DropFiles must be added manually if you're using an event sink object (i.e. methods, not functions). Has that been noted (might want to verify it first)?

I don't know that it makes much sense to document the events on the GuiCreate page.

Ragnar-F:
You're right. I'll add/change this fact about +DropFiles.

Atm I've no better idea where to put them. I just thought that since GuiCreate is quasi pure window-related like the window events (instead of control-related) it would be a good place. Maybe an extra page for them (even though it would be a short page)? But in which folder? misc?

There some more discrepancies, for example the pages GuiControls, TreeView, ListView. Since they have nothing to do with commands/functions anymore, their current location "commands/" is wrong. Principally they're object-related, but the objects folder doesn't feels good ;)

Another consideration would be to create a new folder for gui-related things, for example "gui/" which would be also a good place for a own window event page. But then there're new questions like should we put the Gui/GuiControl object pages in there too, etc.

Lexikos:
Having the Gui controls under commands/ wasn't particularly rational in the first place; they're related to commands, but so is everything else in the documentation.

The pages aren't organised by topic, so gui/ wouldn't fit with the current structure. Of course, the current structure can be changed.

"GuiCreate" is very specific, and if it remains its own page, should only be about that function. The Gui events are also properties of the Gui object, so I guess they belong with the properties and methods. Although they are connected with the third parameter of GuiCreate, there are also properties which control the events (and specific properties for each event). It can also be said that all of GuiCreate's parameters are directly connected with members of the Gui object (Title and Opt).

Ragnar-F:
I moved the Gui events to Gui.OnEvent. I didn't separate them, because of your OnEvent() idea. Are you satisfied?^^

Lexikos:
Yes, that's better.

Lexikos:
Thanks for doing this. It will help the new GUI API into the main v2 branch a little sooner. There's still some changes I want to make before doing that. I've merged the pull request so that it's easier for me to update the documentation as I go.

Ragnar-F:
No problem. Don't forget the sentences and paragraphs which are commented out. I think most of them could be removed easily.

PR #178, opened 2017-04-09
Fix links and Add missing events to A_EventInfo

Ragnar-F:


Lexikos:
There is one link where it appears the only change was to replace `&amp;` with `&`, which GitHub highlights as an error. Why was this change made?

Ragnar-F:
In fact, in this case the ampersand char doesn't matter, because no valid entity can be interpreted after the this character. But for the sake of accuracy, I've changed it.

Lexikos:
I know it works either way. My point was that I was under the impression what you changed it to was *invalid* according to the standard, and since it already worked (and as far as I know, was correct), there was no reason to change it. I asked because I thought maybe you knew something I didn't, and had some reason to change it.

Ragnar-F:
No, no, this change was just a personal interest. I use a self-written script to check the anchors/links, in conjunction with WinHttp.WinHttpRequest.5.1 which is obviously unable to handle links with html entities correctly (.Status returns 404).

Ragnar-F:
You can merge this. I've removed this particular change and rebased the commits.

PR #184, opened 2017-05-28
v2: Clarify Gui._NewEnum()

Ragnar-F:


Lexikos:
Object.Enumerator enumerates key-value pairs. It's not a general enumerator term. COM enumerators return a safe-array element and its type, or an IEnumVARIANT enumerated value and its type.

I don't think it's meaningful or useful to use the term "key-value pair" here. The enumerator just returns the HWND and GuiControl object of each control.

Ragnar-F:
I've changed it to "Returns a new enumerator to enumerate the HWND and GuiControl object of each control inside a GUI."

PR #187, opened 2017-06-26
v2: Add missing StrLen page, adjust code examples to v2 and remove command-specific things

Ragnar-F:
Some questions:

Do we keep the term "command"? Maybe for functions which don't have an explicit return value? Or should it replaced with "function" (which I would endorse personally)?

What about _FunctionName_() outside the code or pre tags. Should the parentheses removed completely, appended to every _FunctionName_ or just appended to _FunctionName_ which are generally used with parentheses (functions with an explicit return value)?


Lexikos:
In general, use "function" unless it's a control flow statement.

No (). Use "the xxx function" if you think it's clearer. "The xxx command" was used pretty often.

PR #188, opened 2017-07-03
Tutorial: Replace * modifier with +

Stevoisiak:
At this point in the tutorial, the `*` wildcard character has yet to be explained.

As a new user reading this tutorial, I was only familiar with the modifier symbols mentioned in _2a. Keys and their mysterious symbols_. [`#`, `!`, `^`, `+`, `&`.]

### DIFF
--- a/docs/Tutorial.htm
+++ b/docs/Tutorial.htm
@@ -350,7 +350,7 @@ Return
 					</p>
 
 <pre>
-:*:ftw::Free the whales
+:+:ftw::Free the whales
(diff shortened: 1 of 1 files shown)


Stevoisiak:
Nevermind. Did not realize [Hotkeys](https://autohotkey.com/docs/Hotkeys.htm) and [Hotstrings](https://autohotkey.com/docs/Hotstrings.htm) had separate modifier characters.

PR #196, opened 2017-09-20
Click: Add commas for consistent syntax

Stevoisiak:
The lack of commas after command parameters is inconsistent with the syntax from other documentation pages. (i.e. `Click 100, 200 right` vs `Click, 100, 200, right`)

While the commas are not *technically* needed, omitting them adds unnecessary ambiguity for novice users. The difference in syntax also makes the transition from MouseClick to Click a bit more confusing.

### DIFF
--- a/docs/commands/Click.htm
+++ b/docs/commands/Click.htm
@@ -19,27 +19,27 @@
     <td>Clicks the left mouse button once at the mouse cursor's current position.</td>
   </tr>
   <tr>
-    <td>Click 44, 55</td>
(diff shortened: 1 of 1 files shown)


Lexikos:
Click does not have syntax consistent with MouseClick. The point of the examples is to demonstrate the syntax, including the fact that it differs from most other commands. The paragraph below the examples table describes something clearly different to a conventional parameter list, though perhaps it would be clearer if it was explicitly compared to other commands.

My interest is low right now, so I don't intend to do anything about this.

PR #199, opened 2017-10-10
 Functions: Clarify that = and := function identically

Stevoisiak:
On the page for [*Variables and Expressions*](https://autohotkey.com/docs/Variables.htm#Variables_and_Expressions), the `=` operator is described as follows.

> The traditional method uses the equal sign operator (=) to assign **unquoted** literal strings or variables enclosed in percent signs... By contrast, the expression method uses the colon-equal operator (:=) to store numbers, **quoted** strings, and other types of expressions.

To prevent confusion, this note clarifies that the two operators function identically when used to define optional parameters.

### DIFF
--- a/docs/Functions.htm
+++ b/docs/Functions.htm
@@ -72,7 +72,7 @@
 <pre>Add(X, Y, Z:=0) {
     return X + Y + Z
 }</pre>
-<p>As of v1.1.09, both <code>=</code> and <code>:=</code> are supported and function <strong>identically</strong> to each other. (ie: both interpret values as an <a href="Variables.htm#Expressions">expression</a>) The latter is recommended for consistency with expression assignments and compatibility with future versions of AutoHotkey.</p>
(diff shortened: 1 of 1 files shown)


Lexikos:
`=` and `:=` in this context are not to be confused with assignment operators. Infact, **neither one** accepts expressions. A function declaration is not an expression; it merely uses syntax resembling a subset of expressions. That being the case, the documentation you quoted from *Variables and Expressions* is not relevant.

Being an imitation of expression syntax, it follows that all strings must be quoted. `=` cannot be used to assign an unquoted string within an expression; why would it here?

The documentation is already clear about whether unquoted strings are allowed: they are not.

> A parameter's default value **must be** one of the following: true, false, a literal integer, a literal floating point number, or a **quoted**/literal string such as "fox" or ""

Perhaps it would be better to replace:
> As of v1.1.09, both = and := are supported. The latter is recommended ...

with:
> := is recommended over = ...

The original point of the paragraph was that `:=` is recommended, **not** that you can also use `=` (which was already explained). It seems backwards to explain that they both accept expressions (putting aside that such explanation is redundant, and actually inaccurate), when `=` has always accepted a quoted string, long before `:=` was supported. In existing scripts with `="..."` it is fairly clear that you've got a quoted string, and in new scripts it shouldn't matter what `=` accepts since you should be using the recommended symbol, `:=`.

Lexikos:
788afdd supersedes this pull request.

PR #202, opened 2017-11-03
Sidebar Fixes

Ragnar-F:


derz00:
image
After switching to German docs with the nifty switch at the top, then back to English, the sidebar is still German...

Ragnar-F:
92d7725 will fix this issue, but it needs to be merged first.

PR #209, opened 2017-11-26
Sidebar Changes/Fixes

Ragnar-F:


derz00:
Any reason the sidebar is a different size for the v2 docs?

Ragnar-F:
I see no difference in size. Could it be that you have set a different zoom level for one of the sites?

derz00:
Sure enough, that's what happened. For some reason I thought my browser (Brave) wasn't saving site-specific zoom levels. And the difference was slight enough that I thought the main page wasn't a different size. 😒 

Thanks

Lexikos:
One flaw with the sidebar that I didn't notice until now is that it is all considered part of the document.  This is a problem because when one tries to search the page (Ctrl+F), one may have to skip through all of the results found in the sidebar.

joedf:
Well that's no good... :( I'll see if there are any workarounds...

joedf:
Here's 2 (working?) techniques:
- using css 'content:' https://stackoverflow.com/a/32398699/883015
- inserting random invisible characters https://github.com/seangransee/Disable-CTRL-F-jQuery-plugin

Ragnar-F:
Using CSS property 'content' is the solution. Thanks for your effort, @joedf.

joedf:
Great! 👍

Lexikos:
Thanks guys, I wasn't aware CSS `content` could be used in that manner.

PR #217, opened 2017-12-12
Sidebar Changes/Fixes

Ragnar-F:


derz00:
I just noticed that when switching between v1 and v2 docs, it goes to `/docs/AutoHotkey.htm` Wouldn't there be a way to go to the equivalent page, if there is one? Or isn't that a desired feature?

Ragnar-F:
The v2 docs also includes structural changes; this means that the counterpart of some pages does not exist, has been removed or renamed (e. g. EnvAdd vs DateAdd), resulting in 404. Right now, I don't have a better idea how to solve this.

Lexikos:
Any idea why navigating to a URL with an anchor, such as [Script File Codepage](https://autohotkey.com/docs/Scripts.htm#cp), in Firefox Quantum (57.0.2) adds three entries to the history (#cp, # and #cp) instead of one? It does not happen in Chrome, nor with the original v1.1.26.01 documentation source.

Ragnar-F:
"Improve scroll position handling when traversing the history" will fix this.

Lexikos:
Thanks. I'll merge it soon.

PR #218, opened 2017-12-13
Forum Suggestions

Ragnar-F:
Arrays.htm: https://autohotkey.com/boards/viewtopic.php?p=188026#p188026
IE/Edge search bug: https://autohotkey.com/boards/viewtopic.php?p=188631#p188631

### DIFF
--- a/docs/misc/Arrays.htm
+++ b/docs/misc/Arrays.htm
@@ -10,21 +10,29 @@
 
 <body>
 <h1>Arrays</h1>
+<p>In AutoHotkey, there are two different types of arrays:</p>
+<ul>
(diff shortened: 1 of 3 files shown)


derz00:
Nice work. I would advise to put _less_ emphasis on the concept of Pseudo-arrays, but that's just my opinion... and would pseudo-arrays work the same in v2? Or work at all?

Lexikos:
I agree with derz00, although my thoughts were more along the lines that psedo-array isn't a type of array; it is a way to fake an array, hence "pseudo". Perhaps there should be something to indicate it is an obsolete concept (not necessarily discouraging its use, but something to show that arrays should almost certainly be used instead).
they are a historical artifact, which implies that

PR #221, opened 2017-12-27
Fix typos

Ragnar-F:


Lexikos:
"Script interacts" was not a typo. It was using "script" in the sense of "code" rather than "a script file".

Adding "s" results in the grammatically incorrect phrase "Scripts interacts". I have changed it to "Scripts interact".

Under the "Methods" section, it was not merely missing a full stop, but the rest of the sentence. ;) I've fixed it, though I feel it still needs work.

PR #223, opened 2017-12-29
Fix small issues

Ragnar-F:


Lexikos:
"When necessary" is not an error, but "When it's necessary" is also fine.

"If var is type" is a type of if statement, not a type of if statements. "If var is type" and "If var in matchlist" are two types of if statement.

"Hotkey labels and hotstrings" was intentional. `#If` applies to hotstrings, not just hotstring labels. `::ftw::for the win` is an autoreplace hotstring but **not** a label. (`#If` really applies to hotkey variants, not the label, which can also be called directly; but that's too much nitpicking for this section. The point of "hotkey *labels*" is that `#If` does not affect the Hotkey command in the same way.)

Subroutines: again, there's a comma because I forgot to finish the sentence, not because I used the wrong punctuation. ;) That's probably where I stopped to write the control flow section in Concepts.htm. I think this sentence was useless anyway.

    -</html>
    \ No newline at end of file
    +</html>

Is there some rule I'm not aware of regarding newlines at the end of a file, or is this just noise?

Ragnar-F:
> Is there some rule I'm not aware of regarding newlines at the end of a file, or is this just noise?

Yes, there is a rule. See https://stackoverflow.com/a/729795.

If I change a file directly on GitHub, this happens automatically due to `git diff`. But also many well known editors like Atom or Vim have this as default.

Lexikos:
I wouldn't call that a "rule", at least not one that's relevant in the context of this project, but at least there's a reason for the noise.

PR #226, opened 2018-01-01
Fix HTML Parsing Error on IE8

### LINKED ISSUE #222: Error KB927917 on XP/IE8 when navigating via Go -> Home/URL
> HTML Parsing Error: Unable to modify the parent container element before the child element is closed (KB927917)

This error appears when one uses Go -> URL... or Home Page on XP/IE8. I'm guessing that it is due to the IE version, not the OS version.

The error dialog shows line 0, and seems to appear after the script executes, not during. I was unable to determine which part of the code causes it.

@Ragnar-F, can you look into this?

If you do any testing, pull down commit 7a05237 first.

Ragnar-F:
This fixes #222


Lexikos:
Thanks. You're a miracle worker.

PR #228, opened 2018-01-03
User Suggestions

Ragnar-F:
Deprecated notes: https://autohotkey.com/boards/viewtopic.php?p=189900#p189900
Help button option number: https://autohotkey.com/boards/viewtopic.php?p=195324#p195324
MsgBox option table: https://autohotkey.com/boards/viewtopic.php?p=195355#p195355
IE8 issues: https://autohotkey.com/boards/viewtopic.php?p=195518#p195518


derz00:
That was a lot of work. You have invaluable skills/dedication.

Lexikos:
Good work, though I disagree with some of the changes.

EnvAdd/EnvSub can't be deprecated until DateAdd/DateDiff are available in v1.

Goto is not deprecated in the sense of the word usually intended with software. It is not obsolete. Unlike most of the others, it will not be removed.

#IfWinActive etc. are not deprecated. Do not recommend #If over #IfWinActive. #If is more flexible, but also slower and has a higher risk of causing input lag. For instance, copy a large file with FileCopy and try pressing an #If hotkey (without a global variant) before it completes. Assuming the FileCopy does not complete within the timeout period, the hotkey will not fire. On some systems, if #IfTimeout is exceeded the OS will also silently remove the hook.

Ragnar-F:
Wow, I didn't know that the disadvantage of #If is so significant.

I've reverted the changes you disagree with.

Lexikos:
A few more:

When I open a page with old and new syntax, I see nothing but the old syntax at first. If I want to refer to the recommended syntax, I have to scroll down. Perhaps they should be swapped around.

> If: Instead, use the specified statement next to each command below [...]

This is a poor recommendation, as these statements are also deprecated. [Language.htm](https://autohotkey.com/docs/Language.htm#if-statement) already recommends avoiding them. (Better integration of this document is needed.)

MouseClick may be preferable over Click in some cases. My current feeling is that there is very little reason to remove it in v2, so I don't consider it "deprecated".

I think it's better without `.methodShort`. In fact, I don't think the current styling of `.methodShort` adds any value over the global styling. A bigger margin between sections might be better, but if so, maybe that should be done for all h2 anyway.

Ragnar-F:
I've changed every point except the last one.

For this style, I took the [jQuery docs](http://api.jquery.com/attr/) as an example. `.methodShort` is most similar to this style, even if the name doesn't really fit in this case.

Of course, this always depends on the personal preferences. I've tried it without this class and come to the conclusion that there is a loss of overview (additionally it looks awful). I've also tried using dividing lines (`<hr>`), which doesn't work either. I don't think a bigger margin between h2s will help much, because it would cause users to scroll more.

Lexikos:
The jQuery docs are much more consistent. Our docs do not use this style for most things. What's the difference between this and every other part of the docs which is split into multiple sections?

> ... there is a loss of overview (additionally it looks awful)

As I said, my personal opinion is that it adds no value (= removing it causes no loss), and it looks worse.

It looks out of place. When I first added it (I think the other styles were different then), I was trying to add visual distinctions between the sections. Now I think it just adds noise, making the page harder to look at. Big colourful headings with (faint) underline is enough everywhere else; it's enough here.

docs/commands/RegDelete.htm
```diff
@@ -13,40 +13,60 @@
...
+<pre class="NoIndent">RegDelete, HKEY_LOCAL_MACHINE, Software\SomeApplication, TestValue</pre>
+
+  </div>
+<h2>ErrorLevel</h2>
 <p><span class="ver">[v1.1.04+]</span>: This command is able to throw an exception on failure. For more information, see <a href="Catch.htm#RuntimeErrors">Runtime Errors</a>.</p>
 <p><a href="../misc/ErrorLevel.htm">ErrorLevel</a> is set to 1 if there was a problem or 0 otherwise.</p>
 <p><a href="../Variables.htm#LastError">A_LastError</a> is set to the result of the operating system's GetLastError() function.</p>
-<h3>Remarks</h3>
+<h2>Remarks</h2>
 <p class="warning"><strong>Warning</strong>: Deleting from the registry is potential dangerous - please exercise caution!</p>
```
derz00: Should be: `potentially dangerous`

derz00: This is directed at line 63. Not sure if that is clear.

Ragnar-F:
"Use relative lengths" fixes #234.

PR #231, opened 2018-01-10
v2: Fix issues

Ragnar-F:


Lexikos:
Some issues I've addressed:

"Function" and "command" are not interchangeable, so take care when replacing one with the other. "Command" could mean WinGet, MouseGetPos, etc. or it could mean a particular instance of calling a command, as in `WinGetPos x, y`. `Run "Notepad"` is not a function; it is a function call, or function call statement. *Run* is a function.

> However, since SendInput is unable to detect a low-level hook in programs other than AutoHotkey v1.0.43+, it will not revert in these cases, making it less reliable than SendPlay/Event.

This must not be removed; it is very important to anyone who is using a *program other than AutoHotkey* which uses a low-level hook.  It is even relevant if some oddball happens to run v1.0.42 alongside v2, because v2 will not detect the low-level hook which the v1.0.42 executable has installed.  It's unlikely that anyone is using this very old version, but in this case only the "v1.0.43+" part can be safely removed.

> In AutoHotkey v2.0 [*was v1.x*], simple arrays and associative arrays are the same thing. However, treating <code>[]</code> as a simple linear array [...] improves the chance of your script working with a future version of AutoHotkey, which might differentiate between simple arrays and associative arrays.

Don't count on it. The point of this paragraph was that they may very well not be the same thing in v2.0*. If it is decided that they will be (which is unlikely, but you couldn't know that, and I won't go into it here), most of the paragraph should be removed from both v1 and v2 documentation.

\* When I say v2.0 (emphasis on the point-oh), I'm often referring to the hypothetical final release, not the current alpha.

> AutoHotkey versions older than v1.1.06 behave as though <code>#InputLevel 0</code> and <code>SendLevel 0</code> are in effect.

This is still relevant to v2. SendLevel can be raised in a v2 script to trigger hotkeys in a v1.1.05 or older script, because they act as though `#InputLevel 0` is in effect. (Also, I noticed the code tags were missing from one file.)

> Warnings are also shown for output variables of functions [*was commands*], but not ByRef parameters.

In this case, "commands" translates "built-in functions called by a function call statement".  I wouldn't expect you to know this.  It is specifically ARG_TYPE_OUTPUT_VAR args which are detected.

One remaining issue (that I don't expect you to address) is that "output variable" or "OutputVar" is not explained in the v2 documentation.  In the v1 documentation it is explained as part of *command syntax*.  It still has a specific meaning but is not really a part of the language anymore, so it may be unclear that it isn't just based on usage (e.g. a ByRef parameter which is used to output a value).

PR #232, opened 2018-01-18
Sidebar Changes

Ragnar-F:


Lexikos:
I think that one of these commits has broken the browser's built-in keyboard shortcut for the address bar (Alt+D in Firefox and Chrome) when viewing the files locally. This is very frustrating.

Ragnar-F:
#235 should fix your frustration (at least in this case) 😉

PR #239, opened 2018-02-18
Content Fixes

Ragnar-F:


Ragnar-F:
`Remove "#IfWinActive" line to avoid confusion`: #IfWinActive without parameter is covered and only relevant in the next examples.

PR #271, opened 2018-04-15
Update startup info for Windows 10

MakotoE:
(In section "How do I put my hotkeys and hotstrings into effect automatically every time I start my PC?")
The current instructions do not work for Windows 10. These new instructions also work for earlier versions.

### DIFF
--- a/docs/FAQ.htm
+++ b/docs/FAQ.htm
@@ -287,9 +287,8 @@ Return</pre>
 <p>There is a folder in the Start Menu called Startup. If you put a shortcut to your script  in that folder, the script will launch automatically every time you start your PC. To create a shortcut:</p>
 <ol>
   <li>Find the script file, select it, and press Control-C.</li>
-  <li>Right-click the Start button (typically at the lower left corner of the screen) and choose &quot;Explore All Users&quot;.</li>
-  <li>Navigate to the Startup folder inside the Programs folder.</li>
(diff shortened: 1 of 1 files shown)


Lexikos:
The old instructions worked on Windows XP. These do not. That's fine.

There are easier ways than what you describe:
  - Run `%APPDATA%\Microsoft\Windows\Start Menu\Programs\Startup` to open the folder.
  - Use `FileCreateShortcut %A_ScriptFullPath%, %A_Startup%` to add the current script to startup. `A_Startup` should always return the correct location, even on Windows XP and older.

MakotoE:
I agree that "Run `%APPDATA%\Microsoft\Windows\Start Menu\Programs\Startup` to open the folder." is easier than what I had said, and should be step 2. However, I am not sure about using `FileCreateShortcut`, since that would involve adding the line into your script and removing it afterwards (or so I presume, I'm not too familiar with AHK yet). Creating the shortcut yourself seems straightforward.

MakotoE:
I understand, I do think the instructions are vital to the FAQ as I came across the FAQ looking on Google for how to make an AHK script load on login.

I suggest these changes: either move the current instructions (in the main fork) to a sub-section, or remove completely, and add instructions that uses `FileCreateShortcut %A_ScriptFullPath%, %A_Startup%\%A_ScriptName%.lnk`.

PR #272, opened 2018-04-16
Redirect missing from repo

ttnnkkrr:


Lexikos:
It was intentional. This is a function of the website, not a part of the documentation. Not that I object to adding it to version control.

What is your reason for adding it?

PR #273, opened 2018-04-19
Sidebar Changes: Settings

Ragnar-F:
Add the possibility to specify the settings the help should be started with by default. For both online and offline help.

With the CHM help, permanent data can unfortunately only be stored in an external file. Cookies and HTML5's LocalStorage are not possible for unknown reasons.

ActiveX is used to save this file. For this to work, ActiveX must be activated (by default this is the case) and the admin rights must be set accordingly. If the requirements are not met, the user is informed and has the alternative of creating the config file manually. See or test the settings page for details.

MS Edge currently has a problem with local pages (not online) in connection with LocalStorage, which means that the help does not work at all. [This issue has already been confirmed](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8816771/) and will soon be fixed by the Edge team.


Lexikos:
I forgot I was going to question this:

> For this to work, ActiveX must be activated (by default this is the case) and the admin rights must be set accordingly. 

That smells like best practices are not being followed. Why does it require admin? Is the configuration file in Program Files?

Ragnar-F:
Yes, I'm not aware of any other solution. I would prefer %Temp%, %AppData% etc. too.

The problem with ActiveX is that by default a security prompt is displayed as soon as ActiveX is used in JavaScript. This happens when the config file is written, which is still bearable. But if you would use ActiveX to read the config file or just to find out the environment variables, you would have to click away this prompt every time you start the CHM file, which is annoying of course. Therefore the only solution is to read the file via HTML script tag, but unfortunately no environment variables, symlinks and so on work this way. Determining these things from the current path of the CHM file without ActiveX is unreliable and tedious. For example: There is no guarantee that %AppData% is on the same drive; different folder structures depending on Windows version; user languages; network drives and so on.

The path where the CHM file is currently located seems to be the only reliable location, since it must exist at this point in time.

PR #289, opened 2018-05-29
Content Changes

Ragnar-F:


Lexikos:
I'm not sure why you think adding ".aspx" to the MSDN links is a "fix". I often omit it because it is both unnecessary and meaningless. (Also, some documents had a suffix like `(vs.80).aspx`, which broke the url bbcode, and it was easier to delete both than just the problematic part.)

But whatever; either way works.

Ragnar-F:
But just for now. As you may have noticed, Microsoft is changing the documentation structure of MSDN, including redirects and removals. In the case of `http://msdn.microsoft.com/en-us/library/ms221702.aspx` when I open this site multiple times sometimes a 404 error shown, sometimes not. That's why I changed this to `https://msdn.microsoft.com/en-us/library/ms221702(v=vs.85).aspx`.

Adding .aspx is just an attempt to prevent future malfunctions. Whether it will work in the end remains to be seen.

Lexikos:
I don't think that makes sense.

`(v=vs.85)` specifies which Visual Studio version or documentation set to pull the document from. If it is omitted, you get a recent version (prior to 2017 it was probably always the latest version). `https://msdn.microsoft.com/en-us/library/6ewkz86d`, for instance, currently shows the VS 2015 version, but adding `(v=vs.120)` will give you the VS 2013 version. No doubt before 2015, it gave the 2013 version by default. Evidently, ms221702 does not exist in the default documentation version, but exists in version "vs.85".

`.aspx`, on the other hand, has no meaning. It makes sense for Microsoft to change the default version of documentation that gets served. It does not make sense for Microsoft to change their URL processing in a way that would break URLs without `.aspx` while not breaking URLs with `.aspx`.

PR #290, opened 2018-05-29
v2: Variables and Expressions - Boolean AND/OR

lvalkov:
I was looking at something tucked away in [v2 Changes](https://autohotkey.com/v2/v2-changes.htm):

> The operators &&, ||, and and or yield whichever value determined the result, similar to JavaScript and Lua. For example, "" or "default" yields "default" instead of 1. Scripts which require a pure boolean value (0 or 1) can use something like !!(x or y) or (x or y) ? 1 : 0.

This behaviour is different to that observed in v1 and isn't necessarily apparent at a glance. I figured it would be a good idea to expand upon it in a place that likely gets more traffic. I added a couple of examples illustrating the usage as well as the type coercion technique.

### DIFF
--- a/docs/Variables.htm
+++ b/docs/Variables.htm
@@ -113,6 +113,8 @@ MsgBox(1.00)  <em>; Shows 1.0</em></pre>
       <p><strong>Unary minus (-)</strong>: Inverts the sign of its operand.</p>
       <p><strong>Unary plus (+)</strong>: <code>+N</code> is equivalent to <code>-(-N)</code>. This has no effect when applied to a pure number, but can be used to convert numeric strings to pure numbers.</p>
       <p><strong>Logical-not (!)</strong>: If the operand is blank or 0, the result of applying logical-not is 1, which means &quot;true&quot;. Otherwise, the result is 0 (false). For example: <code>!x or !(y and z)</code>. Note: The word NOT is synonymous with <strong>!</strong> except that <strong>!</strong> has a higher precedence. Consecutive unary operators such as <code><strong>!!</strong>Var</code> are allowed because they are evaluated in right-to-left order.</p>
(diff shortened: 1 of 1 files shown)


Lexikos:
I think "type conversion" and "pure boolean value" are misleading. The operation does not perform type conversion, but merely produces 0 or 1 (an integer) depending on the input, which can be any type; it is no different from a single logical-not or any comparison operator.

lvalkov:
Okay, how would you rather have it rephrased then?
"Pure boolean value" is lifted from the Changes document. Maybe "Integer boolean value" and then [this](https://github.com/Lexikos/AutoHotkey_L-Docs/pull/290/commits/06f44e4855ad588bf3217c52230b826d3ef2d020#diff-9c739cadd402a42331463638767f21ddR116) becomes "_Use two consecutive **!!** operators to produce an Integer boolean value._"?

Looking back at it, "type conversion" does seem inappropriate. I was probably thinking more along the lines of "coercion", but now I'm not sure this would be a better fit either. Maybe drop it altogether opting for "produces Integer" instead, seeing how a boolean type doesn't even exist.

Lexikos:
If you need a *boolean* value, there's no need for type conversion: just use whatever value and it will be interpreted as either true or false. Doing `!!` is only useful when you specifically need 0 or 1, or something printable (i.e. because objects appear as empty strings when concatenated), so I think describing the actual effect of `!!` (as producing 0 or 1) is more appropriate.

I don't think it needs to be documented specifically. It's something a user can figure out on their own easily enough, if it's what they need. A more general (and therefore more useful) mechanism is ternary: `b ? true : false` or `b ? "true" : "false"` (the latter being useful for string formatting). Again, I don't think the documentation needs to mention that specifically; it's a fairly obvious use of ternary.

PR #303, opened 2018-08-03
v2: Content Changes

Ragnar-F:


Lexikos:
**Decimal and Hexadecimal**

Functions do not (in general) return *decimal* integers. If a numeric string is returned, it is either decimal or hexadecimal; otherwise, it is just an integer. The default display format for numbers is decimal, but this is an attribute of the (implicit or explicit) conversion to string, not the value itself. One could say it is a *pure integer* or *binary integer*, but this is redundant. `Type(the_return_value) = "Integer"`.

For functions that return a "hexadecimal integer", calling it just that may be misleading (outside the context of display format). Considering just the pure type of the value, these are not integers, but strings. To be more precise, it is generally a string of hexadecimal digits either with or without the `0x` prefix.

DateAdd and FileGetAttrib do not return a decimal integer representing a timestamp in YYYYMMDDHH24MISS format. They returns a string of digits in the YYYYMMDDHH24MISS format. This string should not be treated as a number (one should not perform math on it or compare it numerically).

Float's return value, like any other floating-point number in AutoHotkey, is **not a [decimal floating-point number](https://en.wikipedia.org/wiki/Decimal_floating_point)**. It is a [binary floating-point number](https://en.wikipedia.org/wiki/Double-precision_floating-point_format). This affects the precision of operations on the number, as well as its binary representation (as seen when reinterpreting values with NumPut/NumGet or DllCall). Floating-point literals are decimal, as are strings produced by conversion from floating-point numbers. I would just remove the "decimal" qualifier in this case, but perhaps the actual binary format can be documented somewhere.

**Representing**

I think you are overusing the "a *typename* representing the *thing*" pattern. To me, it sounds wrong to say that x *represents* y when x and y are actually the same thing. In particular:

  - Several pages use the phrasing "This function returns a value representing the [actual] value ..." where "a value representing" is *completely* redundant; it requires reading more words before getting to the meaning, without adding anything of use.
  - Some pages use the phrasing "This function returns a *typename* representing the result of *what-the-function-does*." I can see some value in paraphrasing the function's purpose when describing the parameters and return value (and have done so, even though it is redundant), but it can be done in a way that conveys the meaning more immediately, without the self-contained redundancy. The function returns its return value, and the return value is the result of the function; these things need not be repeated.
  - I do not think that "a boolean value representing whether the function was successful or not" is actually an improvement over "1 (true) to indicate success or 0 (false) to indicate failure", which is more precise, more concise and does not require further reading. (However, the "boolean value" link might be useful to someone.)

If you want to "homogenize" the documentation while also clarifying the type of each parameter or return value, I think it is much better to specify the type using separate markup than to force it into a standard template of redundant phrasing. Take jQuery, for example:

> **.is( selection )**
> *selection*
> Type: jQuery
> An existing jQuery object to match the current set of elements against.

... or the Windows API:

> **Parameters**
> *hWnd* [in, optional]
> &nbsp; &nbsp; Type: HWND
> &nbsp; &nbsp; A handle to the window to be tested.
> **Return Value**
> Type: BOOL

For boolean values, there are a few possibilities.
  - "Type: Integer (boolean)"
  - "Type: Boolean (Integer)" seems less correct
  - "True or False"; i.e. list the values (with links), then explain if needed.
  - Just note the actual values (with links) within the explanation.

The Ruby docs put the return value in the heading: e.g. **some_method &rarr; true or false**

ComObjType could have an additional column for "Return Type" (either Integer or String).

**Other**

DllCall: Replacing "DllCall" with "This function" has reduced clarity. (Unlike the other pages, this one is talking about two different functions in the same sentence.)

GetKeyState: While it's true that you can specify a scan code, it will merely be converted to a vk code. The OS does not track the state of each key by scan code. AutoHotkey's keyboard hook might, but GetKeyState does not provide access to that.

GetKeyState: Why remove the phrasing "For keyboard keys and mouse buttons"? (Actually, I notice now it incorrectly said "button", and still uses inconsistent wording: "This function returns", "the return value will be", and "the retrieved value will be".)

PostMessage: "If the string contained by a variable is changed by passing the variable's address to PostMessage," The reason this originally said "SendMessage" is that PostMessage is basically never used this way since it would require more complicated memory management (or using static memory). Posted messages *usually* only have basic numeric parameters (or handles?), not pointers to structs or strings.

RegRead: "**the** actual value of the specified registry **key**" is incorrect; registry keys may have multiple values. The previous text had "**a** value of the specified registry key". Which value is already explained, and the explanation does not need to be repeated here.

SoundGet: "This function returns a value representing the setting of the specified parameters. This value is ..." This uses several words to say almost nothing. I'm not sure if "the setting of the specified parameters" is incorrect or just vague or redundant; the setting is of the sound device, specified by parameters, not of the parameters. The original text plus links to "floating-point number" and "string" would be better.

Ragnar-F:
Thanks for your detailed analysis.

For now, I will use the variant with "Type: ...". It doesn't look nice, but it's still better than the idea of putting the return value in the headline (or the overuse of "representing" 😉). Maybe I'll find a better solution someday.

But how do I handle multiple return types? For example, SoundGet returns either float or string. Should I write "Type: Float, String" here?

> RegRead: "the actual value of the specified registry key" is incorrect; registry keys may have multiple values. The previous text had "a value of the specified registry key". Which value is already explained, and the explanation does not need to be repeated here.

I'm not sure what you want in the end: "This function returns a value of the specified registry key." or "This function returns the retrieved value."?

PR #328, opened 2018-12-04
Update Language.htm

Abounding:
I felt as if the documentation in General Conventions could be a bit more concise, so I gave it a shot.


Lexikos:
I prefer not to sacrifice accuracy or correct grammar for brevity. Also, I try to assume that the reader has zero prior knowledge of other languages.

PR #332, opened 2019-01-03
v2: Remove spaces in ternaries warning

lvalkov:
Removed spaces in ternaries limitation warning and added a link to it. This was supposedly fixed in 2009.

### DIFF
--- a/docs/Language.htm
+++ b/docs/Language.htm
@@ -78,7 +78,7 @@ Methods?
   <li><a href="Scripts.htm#continuation-expr">Continuation by enclosure</a>: A sub-expression enclosed in (), [] or {} can automatically span multiple lines in most cases.</li>
   <li><a href="Scripts.htm#continuation-section">Continuation section</a>: Multiple lines are merged with the line above the section, which starts with <code>(</code> and ends with <code>)</code> (both symbols must appear at the beginning of a line, excluding whitespace).</li>
 </ul>
- 
+
 
 <h2 id="comments">Comments</h2>
 <p><em>Comments</em> are portions of text within the script which are ignored by the program. They are typically used to add explanation or disable parts of the code.</p>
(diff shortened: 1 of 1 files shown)


Lexikos:
No.

There was once a requirement that the ternary operator must be surrounded by spaces, since `?` was originally a valid character for use in variable names. This was changed when objects were added to the language (i.e. at the same time `[]` became operators and became invalid in names).

The documentation is referring to a v2-specific limitation of *expression statements* - see the section heading and explanation below it. In a previous build, `x? y : z` was an expression statement while `x ? y : z` was a parameterless function call (to function *x*) with invalid parameters. 

The limitation was reduced by commit Lexikos/AutoHotkey_L@8e4a12f6 so that `x ? y : z` is valid, but `x + 1 ? y : z` is still a function call (e.g. `MsgBox + 1 ? "positive one" : ""`).

There is still a limitation; just not the exact one alluded to in the documentation.

lvalkov:
I understand that there is a limitation, however, this seems more like a quirk of how commands are parsed rather than a limitation of the ternary operator itself. Going along with your `MsgBox` example, consider the following:

```
CallIfTrue() {
	MsgBox 'Called CallIfTrue()'
	return 'CallIfTrue()'
}

CallIfFalse() {
	MsgBox 'Called CallIfFalse()'
	return 'CallIfFalse()'
}

x(arg) {
	MsgBox 'Called x() with arg: ' arg
}

x := -1 ; Prep. We want 'x + 1' to trigger the else-branch later on.

x + 1 ? CallIfTrue() : CallIfFalse() ; Evaluated as: x(1 ? CallIfTrue() : CallIfFalse())
x + 1? CallIfTrue() : CallIfFalse() ; Also evaluated as: x(1 ? CallIfTrue() : CallIfFalse())

x+1 ? CallIfTrue() : CallIfFalse() ; Evaluated as: (x+1) ? CallIfTrue() : CallIfFalse())
x+1? CallIfTrue() : CallIfFalse() ; Also evaluated as: (x+1) ? CallIfTrue() : CallIfFalse())
```

When I read `"Known limitation: There currently cannot be a space between the variable name and question mark."`, I take it to mean that *this* exact space (the underscore in `x_?`)  **cannot** be there. Unless I misunderstood something, this is demonstrably untrue. It doesn't matter if there is a space or not. It only matters if there are spaces in the condition, however, that's beside the point, as the warning makes no claims regarding this.

Perhaps, it would make more sense to replace the warning with a recommendation to value condition statements surrounded with parens over naked ones, so as to avoid these command parsing pitfalls?

Lexikos:
> this seems more like a quirk of how commands are parsed rather than a limitation of the ternary operator itself. 

It is a limitation of how *statements* are parsed. That is why it is under the heading *Expression Statements*. The context is:

> The following types of expressions can be used as statements:
> ...
> Ternary expressions such as `x? CallIfTrue() : CallIfFalse()`.

I already explained that this specific limitation applies to a previous build, not the current build. The documentation is out of date (inaccurate).

> Perhaps, it would make more sense to replace the warning with a recommendation to value condition statements surrounded with parens over naked ones, 

One could read the very next paragraph after the limitation (in the context of "these expressions can be used as statements"):

> Expressions starting with `(`. However, there usually must be a matching `)` on the same line, otherwise the line would be interpreted as the start of a continuation section.

*Recommendations* can be ignored. If there is no explanation, there will be users asking why one type of ternary expression without parentheses can be used as a statement while another can't, especially since such expressions were used in v1.

lvalkov:
I just saw that you have reworded it in 1ea9a604f07c388d5cd0f2bf6e3dd6f367fa2b92. I'm closing this PR.

PR #333, opened 2019-01-08
Update "Compile AHK" & "Resource Hacker" Links

ashrafhadden:
[Compile_AHK II](http://www.autohotkey.com/forum/topic22975.html) leads to an old archived AHK forum whose download links are all broken. I did some research and found the most active and up-to-date fork of "Compile AHK" (https://github.com/mercury233/compile-ahk).

I also added a hyperlink to the Resource Hacker reference.


ashrafhadden:
Good idea! Seeing as the last commit in [Compile-AHK](https://github.com/mercury233/compile-ahk) was back in 2016 I think it's safe to say its abandoned, so fork away!

PR #337, opened 2019-03-02
Adding VarSetCapacity remarks.

HelgeffegleH:
Reason, this seems to be a recurring source of confusion and surprise.


Lexikos:
I have doubts about this. There are a few scenarios:

  - VarSetCapacity has initialized the variable to be "blank". The value is an empty string.
  - VarSetCapacity has initialized the variable by filling it with a zero byte. The value is an empty string.
  - VarSetCapacity has initialized the variable by filling it with a non-zero byte. The value is a sequence of non-zero bytes.
  - VarSetCapacity has not initialized or modified the variable.

To have data which contains a binary zero (in v1), one must use some other API, like NumPut, DllCall, ClipboardAll or FileRead.  Prior initialization could have been performed by VarSetCapacity or by other means.

It may be more relevant and accurate to document the comparison operators as only operating on numbers and null-terminated strings, not binary data. It doesn't matter where the data came from or how the variable was initialized.

The v2 documentation already has this:

> When comparing strings, these operators compare only up to the first binary zero.

> The == and !== operators can be used to compare strings which contain binary zero. All other comparison operators except ~= compare only up to the first binary zero.

HelgeffegleH:
Hello __Lexikos__ :wave:. Nice to see you are alive, I hope you are also well 👍 .

Good points.

I will remove the first paragraph, although I think it is ok to mention it (but not exactly as I worded it) on this page and on the operators page. The operators pages is ofc the better place to start.

I think the second paragraph is relevant and accurate.

> The v2 documentation already has this:

> When comparing strings, these operators compare only up to the first binary zero.

The recurring misunderstanding seems to be that users do not think of `var` _from_ `varsetcapacity(var, ...)` as a _string_, but as some _binary data_ type, and hence they expect all operators to do what they want and not what the operators are documented to do. That is my impression, I do not know.

Cheers.

Lexikos:
> The recurring misunderstanding seems to be that users do not think of var from varsetcapacity(var, ...) as a string, but as some binary data type, 

Maybe the implication that "a variable which has been initialized with VarSetCapacity" is somehow special would do more to perpetuate the misconception than dispel it.

HelgeffegleH:
I do not know. 

I removed the first paragraph. 

Cheers.

Lexikos:
1) I still do not think these remarks are appropriate for VarSetCapacity, or conducive to a proper understanding of the underlying issue.
2) It seems to be self-contradictory, as in "AutoHotkey does not support [doing x]. To [do x], [how x is done in AutoHotkey]." What I think you're trying to say is that *assignment* does not copy the source variable's entire buffer, but that is because assignment copies the *value* (which can in fact be binary in v1, but only if it came from `ClipboardAll` or `FileRead *c`).

PR #344, opened 2019-03-18
Changing usage of unquoted dllcall arg types to quoted.

HelgeffegleH:
Removing related note and adding quotes where appropriate.

Reason: invalid since: Lexikos/AutoHotkey_L#121


Lexikos:
Something should probably be done about the use of unquoted args in the v1 documentation. It should be officially "deprecated".

PR #348, opened 2019-03-21
Send.htm: Remove extra space from header

stevoisiak:


stevoisiak:
This change can be applied to the documentation for both 1.0 and 2.0.

PR #349, opened 2019-03-23
Update Format.htm

nuj123:
Added a simple code, which seems a bit redundant, but would help out beginners like me understand the command 'Format' a bit more.

### DIFF
--- a/docs/commands/Format.htm
+++ b/docs/commands/Format.htm
@@ -183,5 +183,52 @@ ControlSetText Edit1, %s%
 WinWaitClose
 </pre>
 
+  <pre class="NoIndent">
+  
+<em> ; A slightly more exaggerated example</em>
+one   := Format("{1:010}", 1)	; pad 0s left 
+two   := Format("{1:-010}", 2)	; pad 0s, right 
+three := Format("{1:10}", 3) ; pad spaces, left 
+four  := Format("{1:-10}", 4) ; pad spaces, right
+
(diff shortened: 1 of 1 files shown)


Ragnar-F:
Please check the appearance of your code on the Format page itself.

There are some issues:

- empty lines at the beginning and end of the code
- missing em tags around comments
- by default, code examples in the docs use 4 spaces for one indention; your example uses tabs
- code examples should be fully functional; your example closes immediately after execution (can be prevented with #Persistent or more preferably with MsgBox instead of ToolTip)
- unneeded label "combinations:"
- indexs should be fixed to indexes

It is also a good idea to use a continuation section for this (to save space due to fewer characters). For example:
```
MsgBox,
(
Here it is, with pipes (|) to show borders:
1) |%one%|
2) |%two%|
3) |%three%|
4) |%four%|

5) |%five%|
6) |%six%|
7) |%seven%|

Using "1234.5678":
8) two decimals: %eight%
9) no decimals: %nine%

4 significant figures (sigfigs):
10) 1030.215:	%ten%
11) 001.0302:	%eleven%

12) Combo1:
%twelve%

13) Combo2:
%thirteen%
)
```

nuj123:
Where can I read up on how to format?

Ragnar-F:
Nowhere. Just look at the other code examples (also on other pages) and you will see that there is a certain consistency regarding formatting.

PR #352, opened 2019-03-28
Changes

Ragnar-F:


Lexikos:
In general I do not like "see here" type links, primarily for two reasons:
  - It does not degrade gracefully; i.e. if someone is reading it in a context where the hyperlink is not available, they can't know where to go.
  - It literally does not make sense. "Here" is the place we are at right now, which isn't where the information can be found.

Instead, at least refer to the overall topic or section of documentation - e.g. #If examples.

If you google `don't use click here`, you'll find more reasons. It is apparently a popular blog topic.

Ragnar-F:
I've changed it.

I don't like "see here" either, but my intention was to keep the comments as short as possible.

PR #361, opened 2019-05-08
v2: Changes

Ragnar-F:


Lexikos:
Removal of "The destination directory must already exist." wasn't so much an error as an intentional omission, because I think it does goes without saying (as in, it's not our place to explain how the file system works) and I couldn't be bothered to manually fix that merge conflict at the time.

PR #362, opened 2019-05-08
v2: Add option error handling (for Lexikos/AutoHotkey_L#138)

Ragnar-F:
Docs changes for Lexikos/AutoHotkey_L#138

### DIFF
--- a/docs/commands/TrayTip.htm
+++ b/docs/commands/TrayTip.htm
@@ -73,6 +73,9 @@ HideTrayTip() {
 </pre>
 <p>This may not always work, according to at least one report.</p>
 
+<h2>Error Handling</h2>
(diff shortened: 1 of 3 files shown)


Lexikos:
I would prefer not to say "An exception is thrown if [invalid parameters] are detected" unless there is something more specific to add. This much *should be* universally true, even for functions which never detect invalid parameters, which doesn't make it at all useful. I would prefer not to specify exactly which invalid parameters are detected, as that would require analysis of the source code or extensive testing to determine, and would need more maintenance when validation is improved.

Lexikos:
I see you've also added this to other pages in 8f777759.

In many cases, this behaviour was already present before try/catch was added to the language; you just couldn't *catch* the exception. General conventions regarding error handling of *programmer errors* such as invalid parameters should be documented *somewhere*, but not repeated on every page.

Ragnar-F:
I've reverted 8f77775 in #363.

PR #369, opened 2019-06-01
Left-associativity hint for the exponentiation operator.

lvalkov:
Lost some time debugging otherwise valid code because I assumed it was right-associative.


HelgeffegleH:
There is a mention in the docs,

> Operators of equal precedence such as multiply (*) and divide (/) are evaluated in left-to-right order unless otherwise specified in the operator table.

[src](https://lexikos.github.io/v2/docs/Language.htm#operators)

If this operator was an exception, it would be mentioned. That being said, it is not uncommon for infix notation to be right-associative, as superscript notation is. So I think your suggestion is reasonable 👍 .

Cheers.

lvalkov:
Indeed, there is. Still, I think an explicit reminder would constitute a worthwhile addition, despite the duplication issues that would arise.

[Some observations](https://codeplea.com/exponentiation-associativity-options)

Lexikos:
There are unrelated changes in your pull request that you have not mentioned. This is poor form, even if the changes are trivial and acceptable.

This applies to all versions of AutoHotkey, so why is this PR targeting the v2 branch?

Would it not be better to change the *behaviour* in v2 and then document right-associativity? As far as I can find in the source code, backward-compatibility is the only reason left-associativity is kept. Changing this is a trivial matter of adding 1 to the SYM_POWER entry in the sPrecedence array.

lvalkov:
I hadn't thought about switching `**`'s associativity, but now that you have mentioned it, I reckon bringing it more inline with its mathematical counterpart would constitute a welcome change in the long run. I find right-associativity to be the more intuitive and sensible approach, despite the lack of a clear-cut consensus on how the behavior should be handled. Though I must admit, I am biased towards this.

That said, I am also open to leaving the operator as is, so long as *something is added to the operator's description* to tip off users about this apparent peculiarity. 

The PR is targeting the v2 branch, because that is the version I use exclusively. As far as I can tell, clicking the *Pen* (Edit) prompts you to merge into v2. I simply mimicked that. If there are any guidelines in place dictating how documentation contributions are to be made, point me towards them and I'll make sure to abide by them next time. It seems the unrelated changes consisted of a trailing space that my editor must have gobbled up and a hyperlink I've added. I could revert those if you want me to.

Lexikos:
Looks like Chris anticipated a change in behaviour:

> // On a related note, the right-to-left tradition of something like 2**3**4 is not implemented **(maybe in v2)**.

PR #374, opened 2019-06-05
Amend "conditional" #Include example

lvalkov:
>  Like other # directives, #Include cannot be executed conditionally. In other words, **this example would not work**:
> 
> ```
> if (x = 1)
>     #Include SomeFile.ahk  ; This line takes effect regardless of the value of x.
>     y := 2  ; And **this line would belong to the IF above**, since # directives cannot belong to IFs.
> ```

1. The example wouldn't work only if the attempt to include `SomeFile.ahk` failed.
2. Cannot be asserted with certainty.

### DIFF
--- a/docs/commands/_Include.htm
+++ b/docs/commands/_Include.htm
@@ -42,10 +42,10 @@
 <p>The <em>FileName</em> parameter may optionally be preceded by <code>*i</code> and a single space, which causes the program to ignore any failure to read the included file. For example: <code>#Include *i SpecialOptions.ahk</code>. This option should be used only when the included file's contents are not essential to the main script's operation.</p>
 <p>Lines displayed in the main window via <a href="ListLines.htm">ListLines</a> or the menu View-&gt;Lines are always numbered according to their physical order within their own files. In other words, including a new file will change the line numbering of the main script file by only one line, namely that of the #Include line itself (except for <a href="../Scripts.htm#ahk2exe">compiled scripts</a>, which merge their included files into one big script at the time of compilation).</p>
 <p>#Include is often used to load <a href="../Functions.htm">functions</a> defined in an external file. Unlike subroutine labels, <a href="../Functions.htm">functions</a> can be included at the very top of the script without affecting the <a href="../Scripts.htm#auto">auto-execute section</a>.</p>
-<p>Like other # directives, #Include cannot be executed conditionally. In other words, this example would not work:</p>
+<p>Like other # directives, #Include cannot be executed conditionally:</p>
 <pre>if (x = 1)
-    #Include SomeFile.ahk  <em>; This line takes effect regardless of the value of x.</em>
-    y := 2  <em>; And this line would belong to the IF above, since # directives cannot belong to IFs.</em></pre>
+    #Include SomeFile.ahk  <em>; The contents of SomeFile.ahk will be inserted at this exact position, regardless of the value of x.</em>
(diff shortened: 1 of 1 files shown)


Lexikos:
"This line takes effect" might mean that we check for a previous #include of this file and actually do nothing because one is found, but we do this "regardless of the value of x". "The contents of SomeFile.ahk **will be inserted** at this exact position" seems more precise, but is incorrect. What #include does has already been explained in detail; the point of the example should only be to show that conditional statements have no effect on #include.

I think the `y := 2` part just distracts from the point. The example was probably written with other directives in mind, where the directive itself has no effect on the structure of the code.

lvalkov:
> "The contents of `SomeFile.ahk` **will be inserted** at this exact position" seems more precise, but is incorrect.

In the general case, it may be incorrect. I took the example at face value (to the extent that this is possible, considering an include file is involved). In this specific example, if `SomeFile.ahk` cannot be included, the example would fail to compile. What I have stated is correct.

_"This line **takes effect** regardless of the value of `x`"_ sounds reasonable, however, it is slightly ambiguous. Does it _take effect_ in the sense that the evaluation of `(x = 1)` has no bearing on whether the `#Include` statement itself is _evaluated_ (since the preprocessor performs the evaluation at compile-time)? Does it _take effect_ in the sense that `SomeFile.ahk`'s contents are pasted at this exact position? As you mention, they may or may not end up being included.

My main gripe is with `y := 2  ; And this line would belong to the IF above (1), since # directives cannot belong to IFs. (2)` 

1. Whether or not the line binds to the above `if`-statement hinges entirely upon the successful (as in code is pasted it) inclusion of the file specified and, in turn, its contents. 
2. I'd argue that `#Include` statements, in fact, _can belong_ to an `if`-statement. One could easily import fragments, the inclusion of which, would significantly alter the underlying structure of the source code. If not specifically highlight this very possibility, the example should at least make a mention of it.

Lexikos:
The directive always takes effect, but the effect depends on multiple factors (whether the file exists, the presence of *i, whether the file has been included already). Anyway, I think this issue is separate to the issues you originally commented about.

I'm not sure I understood your original comment correctly;

> The example wouldn't work only if the attempt to include SomeFile.ahk failed.

If the attempt to include SomeFile.ahk *succeeds* but the author intended it to be skipped, the example clearly has not worked. The example has an intended purpose different to the general purpose of the directive.

The structure of the script and virtually all executable lines are irrelevant if the script exits prematurely due to a load time error, such as a syntax error or #include failing.

Only [statements](https://www.autohotkey.com/docs/Concepts.htm#statement) can belong to if statements. Directives are not statements. 

> If not specifically highlight this very possibility, the example should at least make a mention of it.

I disagree. This example is to demonstrate that #include cannot be executed conditionally. That #include can be used to *build* a conditional statement from fragments in multiple files is irrelevant. If you wish to demonstrate something like that, it should be a separate example. I do not think it is worth mentioning; not only because it would be rarely useful, but because one should already realize that it is possible if one understands how #include works at a basic level.

> Causes the script to behave as though the specified file's contents are present at this exact position.

> A script behaves as though the included file's contents are physically present at the exact position of the #Include directive (as though a copy-and-paste were done from the included file). 

One can copy-paste multiple lines together to form a more complex structure, so of course one can do the same with #include.

lvalkov:
I will be closing this as I have no intention of littering the repo with stale PRs. 

The change appears to be inconsequential at this point in time, though I hope it will be revisited eventually.

PR #393, opened 2019-10-23
Changes

Ragnar-F:


Lexikos:
> Clarify #SingleInstance with unknown parameter 

I feel that documenting it in this way would legitimize the use of invalid parameters, which is not intended. I suppose there is value in explicitly permitting `#SingleInstance Prompt`, which is valid in v2 (but not documented), but no other value should be used.

Most other directives do not validate their parameters, even directives that have no parameters; e.g. `#Persistent Off` isn't what it seems.

> Restore more accurate note regarding blocks 

I think this is, at best, no more accurate than the previous wording. Grouping applies to *statements*, not *lines*; e.g. it can be said that the IF below owns one statement, but two lines:

    if x
        while y
            z++

"specified to be executed" alludes to intention, whereas "If an If owns more than one line" seems to be more about the actual interpretation of the code. For instance, the following IF only owns one line, but maybe the author specified both lines with the intention of executing them conditionally.

    if x
        y()
        z()

Neither wording is particularly clear.

Ragnar-F:
I agree with the first one. I have change this commit to make it similar to the v2 counterpart.

The second change is more accurate in terms of multi-statement lines (see [lmstearn's post](https://www.autohotkey.com/boards/viewtopic.php?p=296044#p296044)). Although there are multiple statements, they don't necessarily have to be enclosed in braces. I hadn't considered this when I changed that particular phrase, so I reverted it for now without thinking much about how I could improve it.

Lexikos:
> and <code><em>*/</em></code> is not preceded and followed by characters other than spaces and tabs

This literally does not cover the case where `*/` is *preceded* but not *followed* by non-whitespace characters ("and" should be "or").

Also, the added constraint does not exist. `*/` can be followed by anything that would be valid at the start of a line, including the start of another block comment.

It is also unnecessarily verbose. I prefer @robertcollier4's suggestion, as in 56153fff.

PR #394, opened 2019-10-23
add note that SetTitleMatchMode Regex also affects ahk_exe

mmikeww:


mmikeww:
fixed thanks

Lexikos:
SetTitleMatchMode already notes that it affects ahk_exe, while ahk_exe already documents the behaviour in full detail; in particular, it shows what the regex is being matched against (see emphasis).
> If the RegEx title matching mode is active, ahk_exe accepts a regular expression which must match the **full path** of the process. Otherwise, ahk_exe accepts a case-insensitive **name** or **full path**

I think that "interpreted as exact names or as RegEx patterns" could easily be misinterpreted to mean that the RegEx pattern is matched against the process name.

PR #396, opened 2019-11-15
Update ClipWait.htm

HelgeffegleH:
Floating point remark


Lexikos:
I guess it wasn't obvious enough? 

> Timeout ... can contain a decimal point 
> Specifying 0 is the same as specifying 0.5.

HelgeffegleH:
Hi, sorry for not supplying any hint on why I made the change.

> I guess it wasn't obvious enough?

I thought it was, but at least three users on the forum thought that _Specifying 0 is the same as specifying 0.5_ meant that _0.5 is the minimum timeout value_. So, for example, 0.25 would also mean 0.5. Regardless of what we think about that interpretation, this change also has the little added benefit of hinting about the relation between seconds and milliseconds, which might be helpful to some.

Cheers.

PR #397, opened 2019-11-15
Update ClipWait.htm (2)

HelgeffegleH:
quirk remark

### DIFF
--- a/docs/commands/ClipWait.htm
+++ b/docs/commands/ClipWait.htm
@@ -19,7 +19,7 @@
 <dl>
 
   <dt>Timeout</dt>
-  <dd><p>If omitted, the command will wait indefinitely. Otherwise, it will wait no longer than this many seconds (can contain a decimal point or be an <a href="../Variables.htm#Expressions">expression</a>). Specifying 0 is the same as specifying 0.5.</p></dd>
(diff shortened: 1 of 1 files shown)


Lexikos:
> Due to backward compatibility,
Is it?
```C++
// Waiting 500ms in place of a "0" seems more useful than a true zero, which
// doesn't need to be supported because it's the same thing as something like
// "IfWinExist".  A true zero for clipboard would be the same as
// "IfEqual, clipboard, , xxx" (though admittedly it's higher overhead to
// actually fetch the contents of the clipboard).
```
I don't necessarily agree with the original reason for using 500ms in place of 0, but I don't think a true zero is useful either.

More to the point, I don't see the purpose of the added phrase.

HelgeffegleH:
Hi.

> Is it?

It is if we would change it otherwise.

Anyways, this is related to #396. I theorized that the reason

> Specifying 0 is the same as specifying 0.5.

could be misinterpreted as described in #396,  was because the behaviour seems so senseless to the reader, that it will assume it can't be true literally: And prescribes some other meaning to it which makes more sense to it. This change emphasizes that this a special case which isn't intuitive and that might halt further interpretations of its actual meaning. I think it also implies that it is better to actually specify 0.5, if that is what you want.

Cheers.

PR #403, opened 2019-12-29
clarify DllCall addr lookup optimization

mmikeww:


mmikeww:
Done. Just 'squash' when merging

Lexikos:
Incorrect. The optimization works for **any** dll for which `GetModuleHandle` succeeds at the time of the optimization, regardless of whether it is a "standard dll". It could have been loaded indirectly by another dll that was loaded indirectly by another dll, etc. or even injected by other system software.

PR #405, opened 2020-01-04
[v2] negative Exception.What value also affects Exception.File

mmikeww:


Lexikos:
This should cleanly merge from v1 (#404), since this line is the same.

PR #412, opened 2020-03-08
Add note on hotstring options when specifying both sendmode and key-delay

lenwhite:
`#Hotstring SE k10` will change the send mode, but the key-delay change will be ignored if the current sendmode is SendInput. Therefore, it's necessary to use two lines:

```
#Hotstring SE
#Hotstring k10
```

Probably should fix this as an issue in the code, but settling for documenting it for now.


Lexikos:
What makes you think so? As far as I can tell, it is simply not true.

With `#Hotstring SE k10`, the string "SE k10" is passed to `Hotstring::ParseOptions`. Before the call `g_HSKeyDelay` and `g_HSSendMode` have their default values. After the call, they are 10 and SM_EVENT.

There is no logic within this directive that depends on the current send mode. It just sets some variables. The net result of changing these variables in two lines vs. one is exactly the same.

In general, it is not appropriate to document bugs with wording that implies they are actually valid behaviour. In general, there is an expectation that future releases maintain compatibility with the documented behaviour of previous versions.

PR #422, opened 2020-05-14
Is MsgBox % Format(), not MsgBox Format()

Attrom:


Ragnar-F:
I'm not sure if you are aware of this, but you are currently trying to request a change in the docs of **AutoHotkey v2 alpha**.

In AutoHotkey v2 alpha, all parameters are expressions by default, so a percent sign is not necessary.

Attrom:
I just found out that I have v1 installed, I thought I had v2. Sorry, I'll close the request

PR #428, opened 2020-06-14
Changes

Ragnar-F:


Ragnar-F:
Main reason for the revision of the files in scripts folders: It is no longer necessary to change the code in the AHK file _and_ in its associated page. This means less work (especially in v2, where these files need to be modified frequently due to ongoing syntax changes). Another positive effect: Reduced CHM size. For more details, see "..." next to each commit message.

PR #429, opened 2020-06-14
v2: Changes

Ragnar-F:


HelgeffegleH:
Thank you very much 👍 .

Minor comment regarding the escaping of { for remappings (docs/misc/EscapeChar.htm), you don't get _a literal brace_. Escaping the brace in this context means that it represents the keyboard key `{`.

Minor inconsistency,

```html
<p>A hotkey's function can be called explicitly by the script only if the function has been named. See <a href="#Function">named function hotkeys</a>.</p>
<p>A hotstrings's function can be called explicitly by the script only if the function has been named, see <a href="#Function">Named Function Hotstrings</a>.</p>
```

Cheers.

Ragnar-F:
> Minor comment regarding the escaping of { for remappings (docs/misc/EscapeChar.htm), you don't get a literal brace. Escaping the brace in this context means that it represents the keyboard key {.

Thanks for the info. I've reversed it.

> Minor inconsistency

Fixed.

PR #434, opened 2020-06-22
Changes

Ragnar-F:


Lexikos:
> Except Clipboard, setting these BIVs has no special effect like the other BIVs.

So what if they don't? None of the other BIVs are writable in v1, hence the special emphasis on read/write. I don't really care whether or how it's noted, but I don't think this change makes any sense (although the wording for Clipboard is better now).

ErrorLevel is special (compared to A_Args and other super-global variables) in that it gets saved/restored by the thread interruption mechanism. The dedicated page for ErrorLevel documents this and the fact that you can set ErrorLevel.

By contrast, there's (now) probably no mention anywhere that A_Args can be assigned a value. It may come as a surprise when A_Args in fact does not contain an array, because another part of the script has assigned it a different value. (A minimal correction would be "Contains an array" -> "Initially contains an array".)

Lexikos:
Now that I've looked over #436, I see why you made the change, and concede that it is better this way; as long as there is some indication that each writable built-in variable is writable.

PR #435, opened 2020-06-24
Search Input: mouse paste events are ignored

joedf:
- Events like right-click cut or paste do not trigger the search until a key up event is triggered. This allows for partial cut and paste events to trigger the search.
- Search was triggered needlessly on left/right arrow keys
- Right-click cut clearing or browser-feature 'x' icon does not trigger search anyhow (before nor after this change)
- `oninput` event is sufficient, no need to handle `onkeyup` additionally
- Testing in IE v11.900.18362.0: any double execution seems to be prevented this way too, which seemed to be unique IE11 in my case anyhow.

### DIFF
--- a/docs/static/content.js
+++ b/docs/static/content.js
@@ -521,8 +521,9 @@ function ctor_search()
     // --- Hook up events ---
 
     // Refresh the search list and show color indicator on input:
-    $searchInput.on('keyup input', function(e) {
+    $searchInput.on('input', function(e) {
       var $this = $(this);
+      var prevInput = cache.search_input; // defaults to undefined
       var input = cache.set('search_input', $this.val());
       // if no input, empty the search list, remove color indicator and return:
       if (!input) {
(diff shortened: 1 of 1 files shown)


joedf:
@Ragnar-F Perhaps, you can test this too, just in case.

Ragnar-F:
Unfortunately the keyup event is necessary for IE8 and IE9. IE8 does not support the input event at all and IE9 only partially supports it. So please change `"input"` on line 524 back to `"keyup input"`. Apparently the cause of these issues was `if (e.type != "keyup")` anyway. Thanks for fixing it.

Could you please apply these changes to the index search onEvent (starting at line 451) as well?

> Right-click cut clearing or browser-feature 'x' icon does not trigger search anyhow (before nor after this change)

I can't confirm that. Works for me. If I use these, the search list is empty as intended (IE11.329.19041.0 and any other browser).

joedf:
I see. IE8 and IE9, then `keyup` is definitely required. Thanks for clearing that up for me.
For the clearing, that's very minor and could just be my test or something on my end.

Ragnar-F:
Looks good, thanks.

PR #437, opened 2020-06-30
Merge the two v1.1.33.00 changelogs together

joedf:
Also, avoids any duplicate id. I'm guessing this was possibly updated by a script, if not and separate change logs with the same versioning are intentional, perhaps add a suffix to the IDs ?

Btw, glad this is finally released and all outstanding issues/PRs have been closed. So, sorry for this one...


Ragnar-F:
Many entries also appear twice in the two sections.

joedf:
Good catch, I didn't look so closely. I wasn't sure which entries were preferred for the similar ones, so I opted for the more wordy/descriptive ones.

Lexikos:
It is the entire June 15 section that should have been removed (with the exception of the Ahk2Exe changes I forgot to mention this time, since they aren't included in the commit history), since v1.1.33.00 obviously was not released on June 15. It would have been, and would have had fewer changes, but the signing of the installer failed. I eventually gave up on Certum and released it anyway.

But this works, I suppose.

PR #438, opened 2020-07-01
v2: Changes

Ragnar-F:


Lexikos:
> Adjust #Requires page to v2 

The example will only run on v1, which was the point of the #Requires directive. The added quote marks are not useful unless you also change the version number in the directive. I'll do that.

PR #443, opened 2020-07-14
v2: Changes

Ragnar-F:


docs/commands/WinGetControls.htm
```diff
@@ -69,20 +72,21 @@ <h2>Examples</h2>
...
-        ToolTip "The active window has no controls."
-        return
+WatchActiveWindow()
+{
+    try
+    {
+        Controls := WinGetControls("A")
+        ControlList := ""
+        loop Controls.Length
+            ControlList .= Controls[A_Index] . "`n"
```
Lexikos, line 82:
```suggestion
        for ClassNN in Controls
            ControlList .= ClassNN . "`n"
```

docs/commands/_HotIf.htm
```diff
@@ -150,8 +150,10 @@ <h2 id="Examples">Examples</h2>
 ^Del::Send "^+{Right}{Del}"
 
 ActiveControlIsOfClass(Class) {
-    FocusedControl := ControlGetFocus("A")
-    FocusedControlClass := WinGetClass("ahk_id " FocusedControl)
+    FocusedControl := 0
+    try FocusedControl := ControlGetFocus("A")
+    FocusedControlClass := ""
+    try FocusedControlClass := WinGetClass("ahk_id " FocusedControl)
```
Lexikos, line 156:
```suggestion
    try FocusedControlClass := WinGetClass(FocusedControl)
```

docs/objects/Object.htm
```diff
@@ -235,7 +235,7 @@ <h2>Properties</h2>
 <h2>Functions</h2>
 
 <div class="methodShort" id="Object"><h2>Object</h2>
-<p>Creates an Object from a list of property name, value pairs.</p>
+<p>Creates an Object from a list of property name-value pairs.</p>
```
Lexikos, line 238: The comma was intended to imply that each pair is two parameters, not one (such as `[name, value]`), although maybe not the best idea. (I had been looking at a JavaScript method that takes an *enumerable*/array of "pairs", where each pair is an array.)

docs/commands/WinGetControls.htm
Ragnar-F: Implemented via the latest commits.

docs/commands/_HotIf.htm
Ragnar-F: Implemented via the latest commits.

docs/objects/Object.htm
Ragnar-F: Well, actually it is clarified directly in the syntax box below how exactly the pairs must be specified. I also noticed that there is quite an inconsistency regarding this writing style throughout the docs:

docs\commands\index.htm:
27: list of property name and value pairs
729: list of key-value pairs

docs\commands\NumPut.htm:
53: <em>Type</em>, <em>Number</em> pairs

docs\commands\WinSetRegion.htm:
32: pairs of <strong>X-Y</strong>

docs\objects\Map.htm:
16, 53, 60, 91, 110, 122: key-value pairs

Actually "name, value" should only be marked somehow that name and value belong together. In this case I would prefer to use quotes rather than format tags like code, em, strong etc., because I find them a bit exaggerated. What do you think about the following:

Creates an Object from a list of property "name, value" pairs.

Lexikos: I thought about it while writing it, but honestly couldn't care less. I have no objection to your suggestion.

However, I think the difference between WinSetRegion and the others is mostly due to the actual syntax of the function, not merely a style inconsistency.

Ragnar-F: > I thought about it while writing it, but honestly couldn't care less. I have no objection to your suggestion.

I've changed it.

> However, I think the difference between WinSetRegion and the others is mostly due to the actual syntax of the function, not merely a style inconsistency.

Don't worry. I leave the "inconsistencies" untouched.

PR #449, opened 2020-08-10
Changes

Ragnar-F:


joedf:
@Ragnar-F 

Hey, I was wondering if you could link V2 from the chinese docs:
https://wyagd001.github.io/v2/docs/AutoHotkey.htm

If you select V2 when on english or german, the chinese is not available even though it does exist.

I imagine it's simply adding into, correct? :
https://github.com/Ragnar-F/AutoHotkey_L-Docs/blob/master/docs/static/content.js
```js
'v2': { 'en': 'https://lexikos.github.io/v2/docs/',
        'de': 'https://ahkde.github.io/v2/docs/',
        'zh': 'https://wyagd001.github.io/v2/docs/' } }
```

Ragnar-F:
Yes, it is simple to add, but it requires a little more than what you have provided. I will add it. Thanks for the info.

PR #451, opened 2020-09-16
Add link to WS_EX_APPWINDOW documents

jerrylususu:
Previous links link to `www.microsoft.com`, which isn't very helpful. Now it links to [Extended Window Styles (Winuser.h) - Win32 apps | Microsoft Docs](https://docs.microsoft.com/en-us/windows/win32/winmsg/extended-window-styles).


Ragnar-F:
Yes, the search provided on www.microsoft.com (top right) is becoming less and less useful. The search results are now divided into two tabs: Shop and Explore. And guess which tab is displayed first by default... I wonder where the connection between WS_EX_APPWINDOW and the Tattoo Designs App is...

Anyway, the wording should be adjusted a bit. The purpose of this search is to find other extended styles, not only WS_EX_APPWINDOW. For example:

> For other extended styles not documented here (since they are rarely used), see [Extended Window Styles](https://docs.microsoft.com/en-us/windows/win32/winmsg/extended-window-styles).

jerrylususu:
Changed accordingly. How does it sounds now?

> For other extended styles are not documented here (since they are rarely used), see  [Extended Window Style | Microsoft Docs](https://docs.microsoft.com/en-us/windows/win32/winmsg/extended-window-styles) for a complete list.

Ragnar-F:
Better, thanks, but there is still a typo: "Extended Window Style" should be "Extended Window Styles" (plural form).

jerrylususu:
Nice catch, didn't notice that. Fixed.

PR #453, opened 2020-09-29
Fix comment for #IfWinActive without parameters

ledenis:
It says:
> To turn off context sensitivity for subsequent hotkeys or hotstrings, specify any #IfWin directive but leave all of its parameters blank.

But the comment next to ` #IfWinActive`  without parameters says:
```
; Any window that isn't Untitled - Notepad
```
It should be any window, including the "Untitled - Notepad" one

Fixing this comment.

### DIFF
--- a/docs/Tutorial.htm
+++ b/docs/Tutorial.htm
@@ -234,7 +234,7 @@ return</pre>
 MsgBox, You pressed ALT+Q in Notepad.
 return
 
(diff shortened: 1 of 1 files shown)


ledenis:
nvm I did not see the shortcut is the same (`!q::`) in both, so the comment is actually correct in this case

PR #458, opened 2021-01-06
Clarify what "AltGr" really means

d0ugparker:
"AltGr" reportedly stands for "alternate graving," which is wrong. Wikipedia shows it stands for "alternate graph," or "alternate graphic," as indicated by the Wikipedia reference.


d0ugparker:
Incidental note, the "Hello World" tutorial needs some work to include the different requirements needed to fork. No, it's not a huge priority, and it's not the most exciting thing to be working on, but it's super frustrating to anyone who's new to Git. —Regards.

PR #460, opened 2021-01-23
Addition for when semicolon needs to be escaped

vladdeSV:
> [ Escaping a semicolon ] is necessary *only* if a semicolon has a space or tab to its left. If it does not, it will be recognized correctly without being escaped.

It is valid to remap the <kbd>;</kbd> key. However, to do so it must be escaped, otherwise the whole line will be seen as a comment.

This PR adds a line of text which mentions that a semicolon also needs to be escaped if it is being rebound.

```ahk
`;::x
`; & a::y
```

The sentence is a bit odd, but I also try to take into account that it does not need to be escaped if the rebind has a modifier

```ahk
!;::z ; does not need to be escaped
````

### DIFF
--- a/docs/misc/EscapeChar.htm
+++ b/docs/misc/EscapeChar.htm
@@ -26,7 +26,7 @@
   </tr>
   <tr>
     <td>`;</td>
-    <td>&nbsp;; (literal semicolon). <strong>Note:</strong> This is necessary <u>only</u> if a semicolon has a space or tab to its left, or when is the first character on a line when being remapped. If it does not, it will be recognized correctly without being escaped.</td>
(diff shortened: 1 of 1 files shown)


vladdeSV:
The line of text could be changed to be more similar to how escaping the colon is being described.

> This is necessary only in a hotstring's triggering abbreviation.

Lexikos:
The comment flag (which in v1 can be changed with a directive) never needs to be escaped if preceded by a non-whitespace character. If it is at the start of the line, it is not preceded by a non-whitespace character.

Your wording is okay, except that it (still) does not cover semicolon as a hotkey (which is not remapping).

Lexikos:
It appears that the wording differs between branches. v1 is more accurate:

> It is not necessary to escape a semicolon which has any character <u>other than</u> space or tab to its immediate left, since it would not be interpreted as a comment anyway.

They probably fell out of sync because they were on separate pages due to the removal of the #EscapeChar directive in v2.

vladdeSV:
Ah, I see. Your explanation does cover it better. Nice 👍

PR #464, opened 2021-02-06
Changes

Ragnar-F:


Lexikos (commented):
I think you (or whoever suggested this) missed the purpose of the original text. If it's just about how big a path you can fit in MAX_PATH, that's just 259 characters. The original text was offering a theory as to why the value is 260 characters. (I forget what the source was, and since it doesn't add up to 260, maybe it's wrong.) Take that away, and there's no reason to separate the drive letter from the remainder of the path. In the first place, the path (even an absolute path) might not even have a drive letter.

PR #472, opened 2021-03-08
v2: Changes

Ragnar-F:


Lexikos:
You put code *in* the auto-execute section, and it is executed *by* the auto-execute thread. I suppose the point of changing to "auto-execute thread" was to avoid the misconception that the position within the script is what matters, or that it has to be a specific position.

It seems the wording I've used for most cases like this is `during <a href="Scripts.htm#auto">script startup</a>`, which I think is more straightforward and should be easy to understand even before reading about the auto-execute thread, or threads in general.

PR #477, opened 2021-04-10
v2: Changes

Ragnar-F:


docs/commands/ComObjActive.htm
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
 <title>ComObjActive - Syntax &amp; Usage | AutoHotkey v2</title>
-<meta name="description" content="The ComObjActive function retrieves a running object that has been registered with OLE." />
+<meta name="description" content="The ComObjActive function retrieves a registered COM object." />
```
Lexikos, line 5: Microsoft uses the exact phrase "a running object that has been registered with OLE", so I'm not sure I'd consider this an "issue", but I prefer the more succinct wording anyway.

docs/commands/ComObject.htm
```diff
@@ -34,9 +34,9 @@ <h2 id="Parameters">Parameters</h2>
 
 <h2 id="Return_Value">Return Value</h2>
 <p>Type: <a href="ComValue.htm#Returns">ComValue</a> or ComObject</p>
-<p>This function returns a COM wrapper object of type dependent on the IID parameter.</p>
+<p>This function returns a COM wrapper object of variant type dependent on the IID parameter.</p>
```
Lexikos, line 37: If you look to the line above, you will see the *type* is dependent on the IID parameter, not just the variant type. The omission of "variant" was intentional.

PR #478, opened 2021-04-13
Add documentation of negitive number support

Themis3000:
InStr() supports passing in a negative number for the parameter "occurrence" in order the find the nth occurrence from right to left instead of left to right. I believe that this is a very useful feature and should be mentioned in documentation.

### DIFF
--- a/docs/commands/InStr.htm
+++ b/docs/commands/InStr.htm
@@ -33,7 +33,8 @@
       <p>Regardless of the value of <em>StartingPos</em>, the return value is always relative to the first character of <em>Haystack</em>. For example, the position of "abc" in "123abc789" is always 4.</p></dd>
 
   <dt>Occurrence <span class="ver">[AHK_L 57+]</span></dt>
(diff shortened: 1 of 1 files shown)


joedf:
Neat, I didn't know about this for `occurrence`!

Lexikos:
This feature doesn't exist.

If you want to search right-to-left, you use a negative offset. This can be combined with a positive occurrence number, but specifying an occurrence number less than 1 always produces a result of 0 (not found).

Themis3000:
This feature absolutely works for me, I just discovered something funky about it though. It seems passing in -1 will find the last occurrence *unless* the last occurrence is the last text in the string. However, passing in 0 for occurrence does work as expected for finding the last value.

Here's a video demonstrating the feature: https://www.youtube.com/watch?v=Rlc-oVKdRLM
I am running ahk version 1.1.33.02

Lexikos:
You haven't specified the Occurrence parameter. That's the StartingPos parameter.

PR #482, opened 2021-04-25
v2: Changes

Ragnar-F:


Lexikos:
Given that this appears to be aimed at making things more consistent, a couple of your choices seem strange.

Aside from *Variables.htm*, there were literally only two instances of `Contains \w+` in the documentation before. We aren't describing a variable (which *contains* a value), but a value being passed to a function. Once you define the parameter and it becomes a local variable, your local variable *contains* whatever value it was last assigned, which might not be the value passed by the caller. The value might even be discarded with `*` and never assigned to a parameter, so I think "Contains" is neither meaningful nor literally true. Generally I prefer to just write what the value is, as it was previously.

I found "Return = *Thrown* is [...]" confusing, for both the removal of quotes and code tags and the use of =. I see that "0 = Meaning" was used for some of the parameters, and suppose you've copied that; but numbers are clearly intended as literal values and are easily distinguished from the explanation. There were actually quite a few instances of `>\w+ = `, but with only one having a word on the left (RegEx), and many having a only single word on the right. Most of those are in list items, and are probably good candidates for tables.

Nearly everywhere else that a parameter accepts one or more values out of a set, each literal value is either in a table cell or emphasized in some way (usually with `<strong>`) and followed by a colon, not an equal sign. (But `<strong>` doesn't always indicate a literal value, whereas `"Return"` is clearly a quoted literal string.) Most parameters use `<p>` for each item, or `<br>` for each item in a tight group, rather than having an indented list inside `<dd>`.

You've changed the *Function* section to be more consistent with *Parameters/AddRemove*, but less inconsistent with *Return Value*, which still uses code tags and colons (as it should).

Ragnar-F:
I should have reversed any issue you mentioned. Thanks for the detailed explanation.

PR #493, opened 2021-07-29
add information on how to format function object

BanditTech:
hotkeys with function objects seem to have their own rules, which would be nice to have on this page :)

### DIFF
--- a/docs/Hotkeys.htm
+++ b/docs/Hotkeys.htm
@@ -38,6 +38,10 @@ return</pre>
 <pre>^!s::
 <a href="commands/Send.htm">Send</a> Sincerely,{enter}John Smith  <em>; This line sends keystrokes to the active (foremost) window.</em>
 return</pre>
+<p>To use a Function Object in a Hotkey:</p>
+<pre>BindObject := Func("MyFunction").Bind("ThisValue")
+hotkeyMyLabel := "MButton"
+<a href="commands/Hotkey.htm">Hotkey</a> % hotkeyMyLabel, % BindObject  <em>; Notice the lack of a comma after the hotkey keyword.</em></pre>
 <h2 id="Symbols">Hotkey Modifier Symbols</h2>
 <p>You can use the following modifier symbols to define hotkeys:</p>
 <table class="info">


Lexikos:
This doesn't fit in the "Introduction and Simple Examples" section.
1. Even basic use of the Hotkey command hasn't been "introduced" by that point, and actually isn't covered on that page at all.
2. It's not something that beginners need to know right away, nor is it particularly *simple*.

The Examples section for the Hotkey command does not have an example for associating a function object with a hotkey. Perhaps it should.

> hotkeys with function objects seem to have their own rules, 

The same paragraph which tells you that you can use a function object with the Hotkey command also tells you the rules. It even has a partial example, although it contained an error, which is fixed by f623d4a.

The only rule of real concern is that you must pass a **variable**, not an expression like `Func("...").Bind(...)`.

> Notice the lack of a comma after the hotkey keyword.

It is perfectly acceptable to put a comma before the `%`, or to omit the comma when passing a literal value like `Hotkey MButton, ...`.

There is no need to place the first parameter into a variable; that probably only distracts from the point.

BanditTech:
But in my experience, the only way to get this type of hotkey working, is to remove the comma as I am highlighting.
I may have found a bug

BanditTech:
It might help to clarify that the final example is using the Function Object and include the keyword in the text

BanditTech:
```AHK
BindA := Func("ActionFunction").Bind("AHK is Great!")
BindB := Func("ActionFunction").Bind("So Flexible!")

Hotkey 1, % BindA, On
Hotkey 2, % BindB, On

ActionFunction(Param:=""){
	MsgBox % "You Pressed " A_ThisHotkey " and the function was passed " Param
}
```

An example such as this in the section would be amazing

Lexikos:
> But in my experience, the only way to get this type of hotkey working, is to remove the comma as I am highlighting.

You must be mistaken. In v1, it is never necessary to *remove* the comma immediately following the command name. There are only cases where the comma must be used; for example, `Hotkey =, On` assigns the string `, On` but `Hotkey, =, On` enables the `=` hotkey. The same is true when you use `% BindA`, except that `Hotkey =, % BindA` is an error (missing end `%`).

If there was any case where the Hotkey command required the comma to be removed, it would be a bug. I think it's impossible, though.

PR #497, opened 2021-09-01
Ahk2Exe Updates for 1.1.33.10 & 1.1.34.00

TAC109:
@Ragnar-F 
Could you check these documentation updates, please


Lexikos:
Is the *verbose* in `/silent [<i>verbose</i>]` meant to be literally the word "verbose", or is it to be replaced with a value such as 1? Normally I think italics would indicate a placeholder, and you did call it a *parameter*, but didn't specify what values the parameter can have.

`involking` should be `invoking`.

It's generally not necessary to merge up to the current head as part of a PR whose purpose is to be merged into the current head, and doing so in this case just creates noise in the commit history. There are multiple ways to fix this, but given how little value the commit history for this particular repository has, I don't think we need to go into it.

Ragnar-F:
> It's generally not necessary to merge up to the current head as part of a PR whose purpose is to be merged into the current head, and doing so in this case just creates noise in the commit history. There are multiple ways to fix this, but given how little value the commit history for this particular repository has, I don't think we need to go into it.

@Lexikos, you can bypass this by selecting and using "squash and merge", which will combine all commits from this PR into one commit.

Ragnar-F:
@TAC109, I have created a [pull request](https://github.com/TAC109/AutoHotkey_L-Docs/pull/1) for your branch. Just merge it, then everything should be fine regarding the common style of the docs.

TAC109:
@Ragnar-F 
I've adjusted the description of the `/silent [verbose]` parameter as requested by @Lexikos. Could you give it the one-over please.

docs/Scripts.htm
```diff
@@ -165,8 +165,8 @@ <h2 id="ahk2exe">Convert a Script to an EXE (Ahk2Exe)</h2>
         <td><span class="ver">[v1.1.33+]:</span> Shows the GUI instead of immediately compiling. The other parameters can be used to override the settings last saved in the GUI. <code>/in</code> is optional in this case.</td>
       </tr>
       <tr id="SlashGuiSilent">
-        <td>/silent [<i>verbose</i>]</td>
-        <td><span class="ver">[v1.1.33.10+]:</span> Disables all message boxes and instead outputs errors to the standard error stream (stderr); or to the standard output stream (stdout) if stderr fails. Other messages are also output to stdout. The optional <code>verbose</code> parameter causes status messages to be output to stdout as well.</td>
+        <td>/silent [<code>verbose</code>]</td>
```
Ragnar-F, line 168: The code tag is unnecessary here.

TAC109:
Fixed. Cheers

Lexikos:
> you can bypass this by selecting and using "squash and merge"

@Ragnar-F, as I said, there are multiple ways, and I don't think we need to go into it. My main reason for bringing it up was to avoid such commits in future.

I avoid merging via the GitHub UI, else I would need to ensure GitHub is up to date before merging, and to update my local repository afterward. Instead I automate it locally, keeping my local repository as the authority (only pushing to GitHub and not pulling), and allowing me to review and test the merge using local tools before finally pushing it to GitHub (which closes the PR).

I assume a squash merge via GitHub would have the same effect as an independent commit via git (whether it is based on a squash merge or not); i.e. the pull request status would be *unmerged*, because the head commit of `TAC109:master` has not been merged into `Lexikos:master`. I therefore prefer not to squash, but for this repository, I don't care all that much.

PR #535, opened 2022-02-16
Added Table Of Contents and Object Headers

acr1m:
- Added Table Of Contents to top of page
- InputHook Object section
	- put `<ul>` inside of `<dl>,<dt><dd>` tags to clean it up
- InputHook Object (`<div class="methodShort">`) sections
	- In order to help break up the space...
	- added new sections, Methods, General Properties, Option Properties
	- added `<dl><dd>` tags for each section
	- added `<h2>` section headers in the <dt> tags of the their respective blocks

PS:	It looks like a got rid of a lot of lines, but I actually just "tabbed" them with 2 spaces; as they now sit inside of `<dl><dd>` tags.

### DIFF
--- a/docs/commands/InputHook.htm
+++ b/docs/commands/InputHook.htm
@@ -15,10 +15,59 @@
 <p>Creates an object which can be used to collect or intercept keyboard input.</p>
 
 <pre class="Syntax">InputHook := <span class="func">InputHook</span>(<span class="optional">Options, EndKeys, MatchList</span>)</pre>
+
+<h2 id="toc">Table of Contents</h2>
+<dl>
+<dd>
+  <ul>
+    <li><a href="#Parameters">Parameters</a></li>
+    <li><a href="#stack">Input Stack</a></li>
+    <li><a href="#object">InputHook Object</a></li>
+    <ul>
+      <li>
+      <a href="#InputHookObj_Methods"><u>Methods</u></a>
+      - <a href="#KeyOpt">KeyOpt</a>
+      , <a href="#Start">Start</a>
+      , <a href="#Wait">Wait</a>
+      , <a href="#Stop">Stop</a>
+      </li>
+      <li>
+      <a href="#InputHookObj_GeneralProperties"><u>General Properties</u></a>
+      - <a href="#EndKey">EndKey</a>
+      , <a href="#EndMods">EndMods</a>
+      , <a href="#EndReason">EndReason</a>
+      , <a href="#InProgress">InProgress</a>
+      , <a href="#Input">Input</a>
+      , <a href="#Match">Match</a>
+      , <a href="#OnEnd">OnEnd</a>
+      , <a href="#OnChar">OnChar</a>
+      , <a href="#OnKeyDown">OnKeyDown</a>
+      , <a href="#OnKeyUp">OnKeyUp</a> <span class="ver">[v1.1.32+]</span>
+      </li>
+      <li>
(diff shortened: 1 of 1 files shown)


Ragnar-F:
No offense, but I like the current state better, plus your changes conflict with the general style of the docs:
- Parameters section is always located directly below the yellow syntax box.
- I understand the usefulness of TOCs, but they are indeed omitted on command or function pages because there is no suitable place and for most they are not necessary due to sparse content and consistent structure (parameters, remarks, examples).
- The extra vertical lines in the InputHook Object sections are redundant (add no benefit and interfere with mobile view).
- Instead of `<h2 id="InputHookObj_Methods">InputHook Object - Methods</h2>` use `<h3 id="Methods">Methods</h3>` etc.

In summary, only adding new h3 headers is reasonable, everything else is against the general style or redundant.

acr1m:
Thank you for taking the time to review my submission.

- Noted, on the point concerning TOC consistency and location.
  - Instead, how about small hyperlinks at the end of each "class=shortMethod" that jumps to the InputHook Object section? 
  - (_example below_)
- I didn't think to check compatibility with mobile, great point. 
	- What's the best way to preview that? 
- I'll adjust the headers from `<h2>` to `<h3>` as suggested. 

----

[EndKey](https://www.autohotkey.com/docs/commands/InputHook.htm#EndKey)
----

Returns the name of the [end key](https://www.autohotkey.com/docs/commands/InputHook.htm#EndKeys) which was pressed to terminate the Input.

<pre>KeyName := InputHook.EndKey</pre>

Note that EndKey returns the "normalized" name of the key regardless of how it was written in EndKeys. For example, {Esc} and {vk1B} both produce Escape. [GetKeyName()](https://www.autohotkey.com/docs/commands/GetKey.htm) can be used to retrieve the normalized name.

If the [E option](https://www.autohotkey.com/docs/commands/InputHook.htm#E) was used, EndKey returns the actual character which was typed (if applicable). Otherwise, the key name is determined according to the script's active keyboard layout.

EndKey returns an empty string if [EndReason](https://www.autohotkey.com/docs/commands/InputHook.htm#EndReason) is not "EndKey".

Back to [InputHook Object](https://www.autohotkey.com/docs/commands/InputHook.htm#object)

##

Ragnar-F:
> Instead, how about small hyperlinks at the end of each "class=shortMethod" that jumps to the InputHook Object section?

Nah, it's not that great either. Maybe less noticeable and taking up space, such as in the top right of the box as an icon, single word or small text. Also, for consistency, it would then have to be changed in all docs that use these boxes.

Another solution would be to get the TOC via JavaScript and display it somewhere in the navbar. I'm not sure which place would be best for that either, plus I don't know exactly how to handle existing TOCs.

You see, I'm also thinking about it, but I've never come to a satisfying solution yet. It would be best to drop this change for now.

> I didn't think to check compatibility with mobile, great point. What's the best way to preview that?

Dev tools (F12) → Toggle device emulation (Ctrl+Shift+M) → Refresh page to enable mobile view (F5)

> I'll adjust the headers from `<h2>` to `<h3>` as suggested.

Thanks.

Ragnar-F:
> I'll adjust the headers from `<h2>` to `<h3>` as suggested.

I have done this via PR #527. Also, this PR adds an auto-generated quick reference to the bottom left of the navbar for each page. This should cover the need for more overview. You can find a screenshot in the mentioned PR.

PR #556, opened 2022-06-12
docs(Type): add mention about classes and arrays

rstacruz:
This adds a mention that `Type(Value)` can also work with arrays and objects.

### DIFF
--- a/docs/commands/Type.htm
+++ b/docs/commands/Type.htm
@@ -23,6 +23,11 @@
   <li>Integer</li>
   <li>Float</li>
 </ul>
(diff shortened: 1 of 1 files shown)


Lexikos:
It says "This function returns the class name of Value. For primitive values, ..."
**not** "For primitive values, this function returns the class name of Value ..."

The qualifier "For primitive values" is clearly being applied to the list below, not the statement preceding it.

> One common use 

should read

> One common misuse 

To check *if something is (an) Array*, use `if something is Array`. Using Type will fail to consider any subclass of Array as an Array.

> To check if a value is an instance of a specific class, use the `is` operator.

PR #557, opened 2022-06-14
Update FAQ.htm

Animan8000:


Lexikos:
The addition of Windows 11 is not relevant to the original point, which is not relevant in 2022. I already removed it from the v2 FAQ when it was updated for the new installer. (Of course, v2 doesn't support XP or Vista.)

The original point was that upgrading added UAC into the equation, first because users were migrating from XP (no UAC) to Vista (yes UAC), then because users were changing from a system with UAC disabled (Vista/7) to one where UAC was mandatory (8/8.1).

PR #558, opened 2022-06-17
Update Program.htm

Animan8000:
Removed the old "formatting" (or whatever to call it) in the old installer file name and updated with the new ones, aside that I added the latest version number in them. I suppose line 173 might be debatable, if that original version implemented the functionality (mentioned on line 172)

### DIFF
--- a/docs/Program.htm
+++ b/docs/Program.htm
@@ -153,9 +153,9 @@ else if SubStr(A_LineFile, 1, 1) = "*" &amp;&amp; A_LineFile != "*#1"
 
 <h2 id="install">Installer Options</h2>
 <p>To silently install AutoHotkey into the default directory (which is the same directory displayed by non-silent mode), pass the parameter /S to the installer. For example:</p>
-<pre>AutoHotkey110800_Install.exe /S</pre>
+<pre>AutoHotkey_1.1.34.03_setup.exe /S</pre>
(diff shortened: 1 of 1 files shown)


Lexikos:
Yes, `/E` was added in v1.1.09.03. Please add a version note (see "Restart scripts" for an example of formatting).

joedf:
@Ragnar-F @Lexikos 
I think there's the opportunity here to use a js or even better yet use css-content to insert the current ahk version where it makes sense. 

Something like:
```css
ahk-version::before {
	content: "1.1.34.04";
}
```
```html
<pre>AutoHotkey_<ahk-version/>_setup.exe /S /D=C:\Program Files\AutoHotkey</pre>
```
will become something like:
`AutoHotkey_1.1.34.04_setup.exe /S /D=C:\Program Files\AutoHotkey`

or do something similar but js string replacement/insert. Thoughts? Or maybe this is not worth it or perhaps bug-prone?

Ragnar-F:
It's not worth it for the few lines. Additionally, the CSS approach might not work with older browsers.

joedf:
Agreed. Just a thought.

Lexikos:
Why does it make any more sense to use the current version?

I honestly don't think this topic was worth the time any of us has already spent on it, but I'm kind of curious.

joedf:
Arguably, no benefit. As you've said, not really worth it. More of a thought experiment, to have it become the correct version and formatting "automagically": using either a css-content trick or some js-string replacement.

PR #563, opened 2022-07-05
'than' instead of 'then'

joedf:
Comparison - reported to me by email


Lexikos:
That doesn't make any sense. There is no comparison being made.

`then` is correct, as in sequence. First it sets the font to normal (resetting multiple attributes, as explained in the preceding sentences), and then it sets the font to italic.

joedf:
Whoops! you're right, I miss read it. Sorry about that. 😅

PR #564, opened 2022-07-10
Update Ahk2Exe documentation

TAC109:
Add 'Compression' to index
Add another example to AddResource


Lexikos:
immediatly -> immediately
the windows explorer -> Windows Explorer 

I suppose that a call to SHChangeNotify might supersede the OS-dependent methods of clearing the icon cache. The installer does it (to refresh caching of settings related to file type registrations), and it usually blanks the desktop icons for a moment. Ahk2Exe could do it (if it works), but it might be overkill.
```AutoHotkey
DllCall("shell32\SHChangeNotify", "uint", 0x08000000, "uint", 0, "int", 0, "int", 0) ; SHCNE_ASSOCCHANGED
```

joedf:
Just my "two cents"... I think it's best _not_ to link to the `raw.githubusercontent.*` directly, since the raw links have a history of changing in the past. Additionally the GitHub source view, asides from many other benefits, has syntax highlighting, line numbers, direct line hyperlinking, and already provides the "raw" option just one click away. it's significantly harder to back to the "github view" from the raw link, _i.e._, manually editing the url in the address bar.

TAC109:
@Lexikos 
Typos will be fixed.

I experimented with the DLLCall and, while it worked well, in the end I decided to leave it out.

- Executing it when adding a user icon worked, but if the user removed the reference to the icon the refresh would not then take place, therefore would be inconsistent.

- Always executing the DLLCall seemed to be overkill as you said.

@joedf 
Thanks for your comment. I'll revert this change.

TAC109:
After further consideration I've added a `Refresh Windows Icons` menu item to the `File` menu  of the Ahk2Exe GUI, which will execute the DLLCall. :)

PR #567, opened 2022-08-10
v1 Changes

kevindevm:
added LocalAppData and A_AppDataLocal


Ragnar-F:
I appreciate your effort, but why do you want to add something that doesn't even exist in AutoHotkey or has nothing to do with it? A_AppDataLocal is not a built-in variable, easy to test with `MsgBox % A_AppDataLocal`. LocalAppData is an environment variable, for which Windows is responsible, not AutoHotkey.

If this is a wish of yours, try the forum. The chance that this PR will be closed without a merge is quite high.

kevindevm:
my bad miss read the [forum](https://www.autohotkey.com/boards/viewtopic.php?t=9535) and i did not try before

PR #570, opened 2022-08-23
v1: Changes

Ragnar-F:


joedf:
Oh okay! After reading that commits summary, I thought all instances of `<kbd>` were removed, but really only in some places. Makes sense, because it does have its place in some cases. :+1:

PR #571, opened 2022-08-28
Update Progress.htm

kyklish:
Is it was typo in `Progress`? You use _Param1_ in explanation, but wrote _ProgressParam1_ in description.


Lexikos:
This came from the original documentation by Chris, who isn't around anymore. I assume it was intentional, because parameters of two separate commands are mixed together in one list. ProgressParam1 obviously only applies to Progress, whereas Param1 is less clear.

There's no need to repeat "Progress" every time the parameter is named within that paragraph, given the context. But maybe the inconsistency was unintentional.

PR #576, opened 2022-09-17
Update StrPut.htm

RaptorX:
It was not clear that all parameters (except for *String*) are optional.


Lexikos:
StrPut has two different usages, both shown in the documentation:
```
StrPut String [, Encoding := None] 
StrPut String, Target [, Length] [, Encoding := None]
```
In the second usage, Target **is not optional**. For instance, `StrPut("a", , 0, "utf-8")` is not valid (it currently crashes the program; go figure).

Your changes indicate that Encoding must be omitted to retrieve the required buffer size. This is obviously incorrect, as the required buffer size depends on the Encoding.

PR #578, opened 2022-09-24
Reformat example in WinTitle.htm

Zkaij:
I know this is a small thing...but it really bothered me. Hope you approve. :P

### DIFF
--- a/docs/misc/WinTitle.htm
+++ b/docs/misc/WinTitle.htm
@@ -131,14 +131,12 @@ WinActivate <em>; Use the window found by WinWait.</em></pre>
 }</pre>
 
 <p>The following example returns if the calculator does not exist, otherwise the calculator will be activated and moved to a new position:</p>
-<pre>if not WinExist("Calculator")
-    return
-else
+<pre>if WinExist("Calculator")
 {
     WinActivate <em>; Use the window found by WinExist.</em>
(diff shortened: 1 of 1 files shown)


Lexikos:
Something was lost in translation when @Ragnar-F updated this example to replace the deprecated syntax. The original example was:
```
IfWinNotExist, Calculator
    return
else
{
    WinActivate  ; The above "IfWinNotExist" also set the "last found" window for us.
    WinMove, 40, 40  ; Move it to a new position.
    return
}
```
The point here is not to demonstrate a preferred style of writing code, but to demonstrate how IfWinNotExist sets the Last Found Window. It isn't as obvious as for WinExist because of the negative condition built into the command.

Putting that aside, the text above your version of the example is not suitable:

> The following example returns if the calculator does not exist

The example returns at the same point regardless of whether the calculator exists. There's no longer any purpose in using `return` in the example.

The example above this one uses `return` because originally both examples were parts of a single example.

Zkaij:
Right, took me a minute to figure out the point about the negative condition. Does that mean this example has lost that purpose, since it's now just 'covering' WinActivate again? At the very least, it sounds like the return should be removed from the else branch in the current example, since it also returns in both cases—except that the previous example you sent also has returns in both branches...

Nonetheless, my apologies: I should have paid more attention to the explanation text!
________________________________

PR #580, opened 2022-10-03
Update ControlGet.htm

kaidoloor:
Added explanation how to obtain ClassNN (via ControlGetFocus).

### DIFF
--- a/docs/commands/ControlGet.htm
+++ b/docs/commands/ControlGet.htm
@@ -206,6 +206,7 @@ ChoicePos += 1  <em>; Convert from 0-based to 1-based, i.e. so that the first it
 <p>Unlike commands that change a control, ControlGet does not have an automatic delay; that is, <a href="SetControlDelay.htm">SetControlDelay</a> does not affect it.</p>
 <p>To discover the ClassNN or HWND of the control that the mouse is currently hovering over, use <a href="MouseGetPos.htm">MouseGetPos</a>. To retrieve a list of all controls in a window, use <a href="WinGet.htm#ControlList">WinGet ControlList</a>.</p>
 <p>Window titles and text are case sensitive. Hidden windows are not detected unless <a href="DetectHiddenWindows.htm">DetectHiddenWindows</a> has been turned on.</p>
+  <p>To obtain ClassNN, use instead: ControlGetFocus, some_variable, A</p>
 
 <h2 id="Related">Related</h2>
 <p><a href="Control.htm">Control</a>, <a href="GuiControlGet.htm">GuiControlGet</a>, <a href="ControlMove.htm">ControlMove</a>, <a href="ControlGetText.htm">ControlGetText</a>, <a href="ControlSetText.htm">ControlSetText</a>, <a href="ControlGetPos.htm">ControlGetPos</a>, <a href="ControlClick.htm">ControlClick</a>, <a href="ControlFocus.htm">ControlFocus</a>, <a href="ControlSend.htm">ControlSend</a>, <a href="WinGet.htm">WinGet</a></p>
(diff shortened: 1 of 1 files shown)


Lexikos:
What is it instead of? I don't think your phrasing makes sense.

Inline example code should be enclosed in `<code>` tags.

ControlGetFocus is not a practical way to retrieve the ClassNN in general. Mostly one would use Window Spy or some other tool to determine it, and then write it in the code.

There's already a paragraph talking about ClassNN, but this new one is separated from it by an unrelated paragraph. If there is some reason for this page to point out how to retrieve the ClassNN of the focused control, it would make sense to do so immediately before/after pointing out how to do it with MouseGetPos.

ControlGetFocus isn't listed under *Related*, which is a bit odd given that ControlFocus is.

kaidoloor:
You ask: "What is it instead of?" -- It is an addition, not replacing anything.

You: "ControlGetFocus is not a practical way to retrieve the ClassNN in general. Mostly one would use Window Spy or some other tool to determine it, and then write it in the code." -- I did not find an easy way to query for ClassNN in the code, and had to spend some hours for this. My recomm'd is for using in code. Window Spy is for manual use, not for use in code.

I try to see if I can place and/or word the change better.

kaidoloor:
Note that this has nothing to do with mouse. An example: I need to find out if the user types a hotstring in Comment box of MS Word, or in another window of MS Word. The only way (it seems) to check that is eg. ControlGetFocus, FocusedControlClassNN, A.

Lexikos:
> You ask: "What is it instead of?" -- It is an addition, not replacing anything.

I think you misunderstood my question. My point is that your own use of the word "instead" doesn't make sense:

> To obtain ClassNN, use **instead**: `ControlGetFocus, some_variable, A`

>  I did not find an easy way to query for ClassNN in the code

There isn't one in v1, but there is ControlGetClassNN in v2.

> Note that this has nothing to do with mouse.

Neither does it let you retrieve the ClassNN of any control, unless it is **the focused control**. While some linkage from ControlGet to ControlGetFocus would be justified, I still think your additions do not make sense. ControlGet doesn't retrieve the position of a control, or the text of a control, either; but ControlGetPos and ControlGetText are only listed in the Related section, not explicitly stated in the Remarks.

kaidoloor:
Yes, I misunderstood your question. "Instead" was supposed to mean "instead of ControlGet", as the help page is about ControlGet, but it fails to produce ClassNN. Hence, "To obtain ClassNN, use instead: ControlGetFocus, some_variable, A".

As to "Neither does it let you retrieve the ClassNN of any control, unless it is the focused control." -- this goes without saying. I would assume ControlGet (Retrieves various types of information about a control.) also needs a focus. I want to point users how to obtain classNN, as there is no good helpfile source for that, and people will try to use ControlGet for that.

Lexikos:
>  I would assume ControlGet (Retrieves various types of information about a control.) also needs a focus.

ControlGet retrieves information from a control which you specify. It does not "need a focus". It needs you to specify the control.

ControlGetFocus does not retrieve information from a control which  you specify, and does not allow you to specify a control. It tells you which control is focused. It is not suitable as a method of retrieving a control's ClassNN. There is no "good helpfile source" for obtaining the ClassNN of a control, because there is no built-in method to do so in v1.

> the help page is about ControlGet, but it fails to produce ClassNN. 

ControlGet also "fails to produce" things that are covered by ControlGetPos and ControlGetText, which are more relevant, because they retrieve information *from* a control, unlike ControlGetFocus. It obviously only retrieves what it retrieves, as listed in the sub-commands. It is the same for WinGet, which "fails to produce" even more things that are covered by other WinGet commands.

A more accurate name for ControlGetFocus would have been WinGetFocusedControl.

PR #586, opened 2022-10-31
Update v2 Tutorial's Object Brace Syntax.htm

luigislam:
v2 requires object literal for the Key in the Brace Syntax aka Associative Array. The doc currently displays a String for a Key which would always produce an error when attempted in v2 so I just removed the quotes on the Keys in the doc.


Lexikos:
This is fine, but note at the top of section 7:

> Warning: This section about objects is outdated.

It has not been adjusted for the changes to the object model in v2, such as the base object type being split into Object, Array and Map, and the new distinction between properties and array elements.

PR #590, opened 2022-11-12
Update Send.htm

bo33b:
I don't know if this was intentional, but I find the Descriptions of these 3 keys confusing:

> {Control down} or {Ctrl down}		Holds Ctrl down until {Ctrl up} is sent. To hold down the left or right key instead, use {RCtrl down} and {RCtrl up}.

> {Alt down}		Holds Alt down until {Alt up} is sent. To hold down the left or right key instead, use {RAlt down} and {RAlt up}.

> {Shift down}		Holds Shift down until {Shift up} is sent. To hold down the left or right key instead, use {RShift down} and {RShift up}.

This text implies, for example, that using {RShift down} and {RShift up} has something to do with "holding down the left ... key instead" which it certainly does not.

I propose to remove the references to {... up} in these descriptions, and instead use the relevant L-key-variant names.

### DIFF
--- a/docs/commands/Send.htm
+++ b/docs/commands/Send.htm
@@ -232,7 +232,7 @@
   </tr>
   <tr class="sep_below">
     <td>{Control down} or {Ctrl down}</td>
-    <td>Holds <kbd>Ctrl</kbd> down until {Ctrl up} is sent. To hold down the left or right key instead, use {RCtrl down} and {RCtrl up}.</td>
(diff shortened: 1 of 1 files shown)


Lexikos:
I think it's very obvious what one needs to do, but the wording is ambiguous in the original and inaccurate in yours.

Using {LAlt down} **and** {RAlt down} will hold down the left key **and** the right key, not one or the other.

PR #596, opened 2022-11-27
v1: Move/Rename files for v2

Ragnar-F:
These changes are necessary to get rid of the "commands" folder name, which no longer has any relevance in v2 (in consultation with @Lexikos).

In v1, this is also done to support future v1-to-v2 merges.

Basically, this is the simplest solution. In short, 1) "commands" has been renamed to "syntax", 2) the object files except Functor.htm have been moved to "syntax" and suffixed with "Obj" except Object.htm, and 3) Functor.htm have been moved to "misc". Of course, it could be categorized more strictly (e.g., by gui), but I don't see much additional benefit currently.


Lexikos:
There are also quite a few directives and functions under "commands".

I can see your point about it being simple, but I do not think "syntax" makes sense as a replacement for "commands". Most pages in the documentation have some part about syntax (and Language is primarily about it), meanwhile the pages under "commands" are not primarily about syntax. I would associate "syntax" more with the core language (currently mostly documented in the root), and commands/objects with "library" or "lib". (I didn't suggest "lib" before because of the long association with user-defined function libraries, and because I didn't particularly want something so general.)

> Of course, it could be categorized more strictly (e.g., by gui), but I don't see much additional benefit currently.

I think that without also taking the opportunity to categorize the pages more logically, merely eliminating the term "commands" has too little benefit to justify changing the location of a large number of pages, even if a few simple redirects allow all old links to continue working.

With the division being basically only "syntax", "misc" and *root*, it hardly seems better than dumping everything into one directory.

Ragnar-F:
I like "lib". I created another PR where the files are categorized by topic, inspired by data_toc. See #599. But still I find this variant (lib) the better one. In my opinion, the categorization should be done in content, not by file. From the point of view of a "maintenancer" categorization by topics is less clear than before and looks cluttered. Sure, one could now add an index file to each topic, but is it worth the effort?

Lexikos:
Was there a specific reason for the "Obj" suffix other than a conflict with Func()?

Something about it is bothering me, especially for v2, where classes and functions are in the same namespace, and classes can be *called*.

For Func, I had previously considered merging the two pages.

Ragnar-F:
No, I only added the "Obj" suffix because of this conflict. If the two Func pages are merged, there is nothing against removing the "Obj".

Edit: I just remembered that File.htm could also serve as an overview page for File and Dir functions. But this can be changed later or solved differently, should it ever be implemented.

Edit 2: By the way, @Lexikos, didn't you want to give me writing permissions for the docs, or did I misunderstand that? 😉

PR #603, opened 2022-12-10
v2: Clarify #InputLevel example

zachpoblete:


Lexikos:
The example is written as a single script, and is not intended to be divided. When it runs on its own, it does not work with default settings because SendInput uninstalls the keyboard/mouse hook temporarily (as mentioned in the documentation for SendLevel and perhaps other places).

If *any other script* has a keyboard hook installed, SendInput will revert to SendEvent mode, and the example will work as is.

The original v1 example works as is, because the default mode for v1 is SendEvent.

The better solution is to show that SendEvent should be used explicitly when the script is intending to trigger hotkeys.

PR #610, opened 2022-12-22
ControlFocus remark

kyklish:
Explanation in canges. Below is example to show it and test proposed example.

```AutoHotKey
#NoEnv
#SingleInstance
Gui New, HWndHGui +LastFound
Gui Add, Text,,
(LTrim
    F1: GuiControl Focus Button2
    F2: GuiControl Focus Edit2
    F3: ControlFocus Button2
    F4: ControlFocus Edit2
    F5: WM_NEXTDLGCTL (WS_TABSTOP) Focus Button2
    F6: WM_NEXTDLGCTL (WS_TABSTOP) Focus Edit2
    F7: Send {Tab}
    F8: Test proposed example for Pull Request
    F9: Reload
    You may need reload script to see difference
)
Gui Add, Button, gButton, Button1
Gui Add, Edit,, Text1
Gui Add, Button, gButton HWndHButton2, Button2
Gui Add, Edit, HWndHEdit2, Text2
Gui Add, Button, gButton Default, Default
Gui Show
Return

GuiEscape:
GuiClose:
ExitApp

Button:
    MsgBox % A_GuiControl
Return

F1:: GuiControl, Focus, % hButton2
F2:: GuiControl, Focus, % hEdit2
F3:: ControlFocus,, ahk_id %hButton2%
F4:: ControlFocus,, ahk_id %hEdit2%
F5:: SendMessage, 0x0028, hButton2, True,, ahk_id %hGui%
F6:: SendMessage, 0x0028, hEdit2, True,, ahk_id %hGui%
F7:: Send {Tab}
F9:: Reload
F8:: ;Example
    hWnd := WinExist("A") ; Get HWND of active window
    ControlGet, hWndControl, Hwnd,, Button2, ahk_id %hWnd% ; Get HWND of second button
    SendMessage, 0x0028, hWndControl, True,, ahk_id %hWnd% ; 0x0028 is WM_NEXTDLGCTL
Return
```

### DIFF
--- a/docs/lib/ControlFocus.htm
+++ b/docs/lib/ControlFocus.htm
@@ -44,6 +44,10 @@
 <p>To improve reliability, a delay is done automatically after every use of this command. That delay can be changed via <a href="SetControlDelay.htm">SetControlDelay</a>.</p>
 <p>To discover the ClassNN or HWND of the control that the mouse is currently hovering over, use <a href="MouseGetPos.htm">MouseGetPos</a>.</p>
 <p>Window titles and text are case sensitive. Hidden windows are not detected unless <a href="DetectHiddenWindows.htm">DetectHiddenWindows</a> has been turned on.</p>
+<p>Setting keyboard focus to target control has different result, then switching focus to same control using <kbd>Tab</kbd> key on keyboard. By default, <em>Button</em> control with keyboard focus can be pressed by <kbd>Space</kbd> key, <em>Edit</em> control has input caret in it. Switching keyboard focus with <kbd>Tab</kbd> key to <em>Button</em> control (manually or by <a href="Send.htm">Send</a> command) additionally changes behavior of <kbd>Enter</kbd> key (it begins point out to focused <em>Button</em>, not to <a href="GuiControls.htm#DefaultButton">default</a> <em>Button</em> in window, if its present); to <em>Edit</em> control - additionally selects all text in it. <em>Tab</em> behavior can be achieved by sending message to window via <a href="../misc/SendMessage.htm">SendMessage</a> command like in this example:</p>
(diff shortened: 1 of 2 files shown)


Lexikos:
I found your wording hard to follow, even already being familiar with the issue you are trying to describe.
  - ", then" should be "than".
  - I'm not sure why you bring up Edit controls or the caret. It seems to be beside the point, unless you were going to also mention that WM_NEXTDLGCTL will select the control's text.
  - "it begins point out to focused Button" makes no sense to me, but knowing what effect the Tab key has, I assume you are trying to say something like "it will activate the focused Button".

But thanks anyway. I understand that English isn't the easiest language to master.

It is not that setting focus to the control has a different result, but that pressing Tab has **additional** effects beyond just changing the focus.  The documentation for WM_NEXTDLGCTL is very clear about the additional effects:
> This message performs additional dialog box management operations beyond those performed by the [SetFocus](https://learn.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-setfocus) function. **WM_NEXTDLGCTL** updates the default pushbutton border, sets the default control identifier, and automatically selects the text of an edit control (if the target window is an edit control).

The change of default button can be observed by which button has the (usually blue) highlight. This is independent of which control has focus (sometimes indicated by a dotted rectangle).

Pressing Tab is not the only way to apply these effects. Clicking on a Button will also both place the focus on it and temporarily set it as the default button, so pressing Enter will activate it. Activating another window and then reactivating the GUI may also do it.

I have updated the documentation to clarify these issues, and incorporated a version of your example. Note that comments in examples should be enclosed in `<em>` for cases where the syntax highlighter does not execute. I opted to replace `ahk_id` with the Last Found Window to avoid repetition and better focus on the point of the example.

kyklish:
Thanks.

PR #611, opened 2022-12-25
Add @Ahk2Exe-Set example

pizzashi:
I think this example is particularly helpful, since it's not supported by `;@Ahk2Exe-SetProp`. The example changes the compiled script's displayed name in Task Manager.

From https://learn.microsoft.com/en-us/windows/win32/menurc/versioninfo-resource

FileDescription - File description to be presented to users. This string may be displayed in a list box when the user is choosing files to install—for example, Keyboard Driver for AT-Style Keyboards. This string is required.

Example demonstration:
```
;@Ahk2Exe-Set FileDescription, MyTestApp
#Persistent
```
Result:
image

### DIFF
--- a/docs/misc/Ahk2ExeDirectives.htm
+++ b/docs/misc/Ahk2ExeDirectives.htm
@@ -362,6 +362,8 @@ MsgBox This message appears in both the compiled and uncompiled script
   <dt>Value</dt>
   <dd>The value to set the property to.</dd>
 </dl>
+<p><strong>Example 1:</strong> This example changes the compiled script's displayed name in Task Manager:</p>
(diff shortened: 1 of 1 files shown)


Ragnar-F:
But... it is supported by SetProp. `;@Ahk2Exe-Set FileDescription, MyTestApp` is the same as `;@Ahk2Exe-SetDescription MyTestApp`. I would just put this info in the description of the Description property, without example, and add "On Windows 8 and higher, ..." (when the new Task Manager was introduced), since v1 supports Windows XP and higher.

pizzashi:
Thanks, I can't believe it's right under my nose this entire time. I guess `Description` instead of `FileDescription` misled me a bit.

PR #622, opened 2023-02-02
Built new search index

white-gthb:


Ragnar-F:
It is not necessary to rebuild the search data so often. This is done automatically with a new release or only if there are problems.

white-gthb:
How are people who update the docs by getting it from GitHub supposed to know they need to rebuild the search index?
Perhaps a readme can be added?

Ragnar-F:
If it were mandatory for contributors, it would be written somewhere. Otherwise, if one knows about build_search.ahk at all, one can assume that the maintainers of these docs care about it, in the sense of whether it is worthwhile to rebuild the search data outside of the automatic rebuild.

white-gthb:
People who update the docs by getting it from GitHub are not necessarily contributors. They are simply users. It seems to me that for them it is worthwhile to rebuild the search data whenever it is outdated.

For contributors this is also important when they want to test the Help after they made changes to it. (as I found out)

Are changes on GitHub automatically published on https://www.autohotkey.com/docs/v2/? Is the search data automatically rebuilt when this happens?

Ragnar-F:
On GitHub, a contributor is a user who has previously committed to a repository. That's you, as seen in your posts at the top right.

If you want to test the search after you have changed something, do it locally. You don't have to commit it.

Changes on GitHub are automatically published. I have already told you when the search data will be rebuilt. It is not necessary to rebuild the search data for every small change. That would clutter up the commit history. Please leave this decision to the maintainers and collaborators (that's me, for example).

white-gthb:
>  It is not necessary to rebuild the search data for every small change.

Well, the changes go live, accessible via AutoHotkey Dash. So there is something to say for a fully working search at all times. It doesn't have to be committed, but the search data on the website should be up-to-date.

> Please leave this decision to the maintainers and collaborators (that's me, for example).

There seems to be some confusion because I have no problem with that. That's why I suggested a readme file for users and contributors who want to download the latest version and use it locally.

PR #628, opened 2023-02-16
Update ListView and TreeView

RaptorX:
Reflect alternate syntax so it can be found on the search function

### DIFF
--- a/docs/lib/ListView.htm
+++ b/docs/lib/ListView.htm
@@ -28,13 +28,9 @@
 <p>A ListView usually looks like this:</p>
 <img src="../static/ctrl_listview.png" alt="ListView" />
 <p>Though it may be elaborate, a ListView's basic features are easy to use. The syntax for creating a ListView is:</p>
-<pre class="Syntax" id="GuiAdd">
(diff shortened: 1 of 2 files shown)


Ragnar-F:
Could you please do it as shown on the [GUI Control Types](https://www.autohotkey.com/docs/v2/lib/GuiControls.htm) page, i.e. 2 syntax boxes instead of one? And take care that `id="GuiAdd"` is allowed to be present only once.

RaptorX:
Sure, will do.

RaptorX:
> Could you please do it as shown on the [GUI Control Types](https://www.autohotkey.com/docs/v2/lib/GuiControls.htm) page, i.e. 2 syntax boxes instead of one? And take care that `id="GuiAdd"` is allowed to be present only once.

done, please review.

Ragnar-F:
Please remove the line breaks within the pre elements. Some older browsers interpret them as empty lines. Also, both "Or:" must be enclosed by p tags.

Ragnar-F:
Thanks

PR #632, opened 2023-02-26
Update GuiControls.htm

iPhilip:
Added note to make sure that the `-Theme` option is included in order for the status bar's color change to take effect.


iPhilip:
Thank you.

PR #633, opened 2023-03-01
Update TreeView.htm

iPhilip:
The word "starting" in "starting state" confused me so I am suggesting "desired" as an alternative.


Ragnar-F:
Why is it confusing? This is the state the item will initially have.

iPhilip:
> This is the state the item will initially have.

That's not my experience. Here is some code that demonstrates that the state represents the final or desired state.

``` ahk
MyGui := Gui()
TV := MyGui.Add('TreeView')
ParentID := TV.Add('Main Item')
TV.Add('Sub Item', ParentID)
MyGui.Show()

F3::
{
   static State := false
   State := !State
   TV.Modify(ParentID, 'Expand' State, 'Main Item' (State ? ' (expanded)' : ''))  ; The variable State represents the final state.
}
```

Ragnar-F:
This option is listed under Add, not Modify, so "starting state" makes more sense in this case. The logical conclusion that it is not the "starting state" in conjunction with Modify should be obvious to most readers.

iPhilip:
Thank you. I agree with your decision.

PR #635, opened 2023-03-08
Fix Add Method

RaptorX:
Includes a brief description of what happens on failure similar to v1 and based on this conversation:

https://www.autohotkey.com/boards/viewtopic.php?p=511167&sid=5bf715b8f4a9fa233dde6865d36ca766#p511167


Ragnar-F:
This is already fixed. See https://github.com/AutoHotkey/AutoHotkeyDocs/commit/9866788f6fcedb34af75cba7445f340741c32c84 and https://www.autohotkey.com/boards/viewtopic.php?f=86&t=114801.

PR #649, opened 2023-06-01
Add example case for sending an expression that contains a variable

ystaggl:
Added an example for sending an expression that contains a variable


Ragnar-F:
"Sends an expression" will cause confusion because everything in v2 is expression syntax (there are a few exceptions such as directive parameters). If you remove everything regarding expression, I'm willing to merge this. An example for variable usage makes sense, especially for beginners.

ystaggl:
Done! Thanks for your input.

PR #652, opened 2023-06-11
Add known limitation to Gui.Call() method

RaptorX:
As discussed in [this thread](https://www.autohotkey.com/boards/viewtopic.php?p=525546#p525546) there are some limitations that cannot be properly documented but that is good to have at least a warning about it when creating a GUI that is a child of an external window.

### DIFF
--- a/docs/lib/Gui.htm
+++ b/docs/lib/Gui.htm
@@ -78,9 +78,11 @@
 <p>Creates and returns a new Gui object.</p>
 <pre class="Syntax">MyGui := <span class="func">Gui</span>(<span class="optional">Options, Title, EventObj</span>)</pre>
 <dl>
-  <dt>Options</dt><dd>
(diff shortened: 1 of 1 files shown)


Ragnar-F:
The current location for the limitation note is not well-chosen. Better to add this after the other known limitation right at the [+Parent option](https://www.autohotkey.com/docs/v2/lib/Gui.htm#Parent), pluralize "Known limitation", change "property" to "option", and insert `via <a href="#Call">Gui()</a>` between "window" and "if".

Does this limitation also apply to v1?

Ragnar-F:
@RaptorX, will you do that for v2, please? Otherwise, I'll have to close this PR and make the change myself.

RaptorX:
Sorry I have been away for a little while.
I'll add it tonight.

PR #664, opened 2023-09-07
Clarify SoundSetVolume offset behavior

G33kDude:
See https://www.autohotkey.com/boards/viewtopic.php?f=86&t=121184


Ragnar-F:
Sorry, seen too late. I have already updated the docs, but thanks anyway.

PR #670, opened 2023-11-02
Update Functor.htm

iPhilip:
I expanded the [BoundFunc Object](https://www.autohotkey.com/docs/v2/misc/Functor.htm#BoundFunc) example to demonstrate calling a BoundFunc object with the first and third parameter of `RealFn`. I had mistakenly tried to do so with `fn(2, , 0)`. This change could help others who are similarly confused.


Ragnar-F:
`fn(2, 0)` shows "2, 1," instead of "2, 1, 0", so `IsSet(c) ? ", " : ""` should be changed to `IsSet(c) ? ", " c : ""`.

iPhilip:
Thank you for catching that! I corrected the entry.

PR #672, opened 2023-11-07
Update Process.htm

icc2icc:
encapsulated as a function

### DIFF
--- a/docs/lib/Process.htm
+++ b/docs/lib/Process.htm
@@ -189,5 +189,22 @@ for proc in ComObjGet("winmgmts:").ExecQuery("Select * from Win32_Process")
 Gui, Show,, Process List</pre>
 </div>
 
+<div class="ex" id="ListFunc">
+<p><a class="ex_number" href="#ListCom"></a> encapsulating in a function & adding closeAll</a>.</p>
+<pre>
+process( "CloseAll","notepad.exe")
+process(SubCommand,PIDOrName,Value:=""){
+	if (SubCommand=="CloseAll"){
+		Loop {
+			process( "Close",PIDOrName)
+		}until !process( "Exist",PIDOrName)
+		return 	
+	}
+	Process,% SubCommand,% PIDOrName, Value
+	return ErrorLevel
+}
+</pre>
+</div>
+
 </body>
 </html>


Ragnar-F:
I will merge this, but with changes, because the code can be simplified significantly and is therefore better placed in the Close section. Encapsulating as a function is rather unnecessary, because there are already v1 libs that provide commands as functions, and there is v2.

Some notes for your function if you want to continue using it: Compared to the command, the function's PIDOrName is not optional for "Exist" and "CloseAll" is case-sensitive due to `==`.

icc2icc:
Thanks for your fix.
Unfortunately, your code didn't work. It only worked when adding a sleep
function.
 (due to the impact looping too fast,so it caused Windows to respond with
error level 0,even when there was still such a process name)
 When I added* sleep, 20*, it worked on two windows, but when I tried it on
multiple windows, it got stuck somewhere.
VS my code that checks for existence in each loop, which actually consumes
more time and resources, but it's at least reliable.
so the most closes to your code that didnt include the function would be


* Loop{ Process, Close, notepad.exe Process, Exist, * *notepad.exe*
* }Until not ErrorLevel*

Also if you can please explain what you meant with *the function's
PIDOrName is not optional for "Exist"*
Thanks

icc2icc:
Thanks for your fix.
Unfortunately, your code didn't work. It only worked when adding a sleep
function.
 (due to the impact looping too fast,so it caused Windows to respond with
error level 0,even when there was still such a process name)
 When I added "**sleep, 20**", it worked on two windows, but when I tried it on
multiple windows, it got stuck somewhere.
VS my code that checks for existence in each loop, which actually consumes
more time and resources, but it's at least reliable.
so the closest to your code (that didnt include the function) in a working example would be


```
Loop{
  Process, Close, notepad.exe
  Process, Exist, notepad.exe
 }Until not ErrorLevel

```
Also if you can please explain what you meant with 

> the function's
> PIDOrName is not optional for "Exist"

Thanks

Ragnar-F:
You're right. Too bad it's not that easy. With Notepad processes it worked (at least for me), but with other processes it may fail. I've changed the example accordingly. Thanks for testing.

> Also if you can please explain what you meant with

For `Process, Exist`, PIDOrName can be blank or omitted to retrieve the script's own PID. PIDOrName cannot be omitted for your process() function (i.e. `process("Exist")`), as the second parameter is mandatory, but it can at least be blank (an empty string).

white-gthb:
Comment about code formatting:
Shouldn't the opening bracket of the loop block be on a new line, so it is consistent with Example 4 and with the examples you see on the help page of the Loop command?

Ragnar-F:
If it does not interfere with clarity, the OTB style is allowed to be used, but it should follow the usual usage in the docs, as shown e.g. on the Loop page (i.e. the opening brace should be preceded by a space).

PR #675, opened 2023-11-21
Added missing " in example.

marska:
Added missing " in example.


Ragnar-F:
Quotes are not missed there. This is v1 code. For v2 docs, see https://www.autohotkey.com/docs/v2/Program.htm

PR #676, opened 2023-11-22
Update Objects.htm

iPhilip:
Clarifying the behavior of the `__New` method when constructing an instance of a class in the [Construction and Destruction](https://www.autohotkey.com/docs/v2/Objects.htm#Custom_NewDelete) section. [[Forum post](https://www.autohotkey.com/boards/viewtopic.php?p=548057#p548057)]


Ragnar-F:
It is up to @Lexikos to decide if this is worth merging.

iPhilip:
Sounds good.

PR #685, opened 2024-03-12
Mention uiaccess

tamo:
If there is an always-on-top UWP window, you need UIAccess to make another window *really*-always-on-top.

For example, run [Movies&TV](https://apps.microsoft.com/detail/9wzdncrfj3p2) (UWP app) in full-screen mode on one monitor. It is always-on-top.
Then move another always-on-top window (of AHK or anything) from another monitor to Movies&TV. It won't appear on top, *even with admin privs*, if it doesn't have UIAccess.

Neither WinSetAlwaysOnTop nor WinMoveTop can bring it up to the top of such UWP windows without UIAccess. Only UIAccess helps in this case.

https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-securityoverview says
> an assistive technology application with these privileges can run as the topmost application in the z-order at any time, meaning that an assistive technology application can be visible and available whenever the user needs it.

### DIFF
--- a/docs/lib/WinSetAlwaysOnTop.htm
+++ b/docs/lib/WinSetAlwaysOnTop.htm
@@ -43,6 +43,7 @@
 <p>An <a href="Error.htm#OSError">OSError</a> may be thrown on failure.</p>
 
 <h2 id="Remarks">Remarks</h2>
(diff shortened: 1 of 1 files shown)


Ragnar-F:
I've just noticed that at the top of the page it says "except other always-on-top windows". So I guess the mention is actually redundant. I'll revert the commit for now. Apparently you can just work around this with UIAccess.

PR #686, opened 2024-03-16
Advise not to make off-by-one errors

tamo:
Though you can use (Top, Left) as the upper-left pixel, you cannot use Bottom or Right as the maximum value within the monitor.
image


Ragnar-F:
There is no need to mention this, you can quickly find out for yourself. Also, this information is not relevant for single monitor environments.

PR #691, opened 2024-04-24
Update File.htm

iPhilip:
Added a comment to document the use of this method to flush the write buffer. Reference: [Example #3](https://www.autohotkey.com/docs/v2/lib/FileOpen.htm#ExStreams).

### DIFF
--- a/docs/lib/File.htm
+++ b/docs/lib/File.htm
@@ -55,7 +55,7 @@
   <dt>Characters</dt>
   <dd>
     <p>Type: <a href="../Concepts.htm#numbers">Integer</a></p>
-    <p>If omitted, the rest of the file is read and returned as one string. Otherwise, specify the maximum number of characters to read. If the File object was created from a handle to a non-seeking device such as a console buffer or pipe, omitting this parameter may cause the method to fail or return only what data is currently available.</p>
+    <p>If omitted, the rest of the file is read and returned as one string. Otherwise, specify the maximum number of characters to read. If the File object was created from a handle to a non-seeking device such as a console buffer or pipe, omitting this parameter may cause the method to fail or return only what data is currently available. If zero, the method flushes the write buffer.</p>
   </dd>
 </dl>
 <h4 id="Read_Return_Value">Return Value</h4>


Ragnar-F:
Read does not have to be zero to flush the write buffer. Furthermore, if I interpret the internal AutoHotkey code correctly, Seek, Length (when set) and Handle can also flush the write buffer in between.

Also informative: https://www.autohotkey.com/boards/viewtopic.php?t=122037

iPhilip:
Thank you for your comment. I was aware of geek's post, which I read when I faced a similar issue. As you stated, it's very informative.

As I read your comment, I realized that I failed to communicate the reason for the edit. I agree that the write buffer is periodically flushed during file operations and then flushed again when the file object is closed via `FileObj.Close()`. There are instances when a user might want to manually flush the buffer, such as [Example <i>#</i>3](https://www.autohotkey.com/docs/v2/lib/FileOpen.htm#ExStreams) in the [FileOpen](https://www.autohotkey.com/docs/v2/lib/FileOpen.htm) documentation page. I found no explicit statement in the documentation, other than the example, on how to do that. That was the reason for my original PR.

After reading your comment, I ran through a number of tests to document which File Object methods and properties flush the write buffer. Overall, they match your observations.

Here is my test script:

```AutoHotkey
#Requires AutoHotkey v2.0

s := '1234567890'
Buf := Buffer(1,0)

; Filename := '*'
Filename := 'foo.txt'
FileObj := FileOpen(Filename, 'w')
FileObj.Write(s)

; FileObj.Read(0)  ; Flushes the buffer - argument can be omitted or any integer other than -1 [Ref 1]
; FileObj.Write(s)  ; Doesn't flush the buffer
; FileObj.ReadLine()  ; Flushes the buffer
; FileObj.WriteLine(s)  ; Doesn't flush the buffer
; FileObj.ReadInt()  ; Flushes the buffer
; FileObj.WriteInt(0)  ; Doesn't flush the buffer
; FileObj.RawRead(Buf)  ; Flushes the buffer
; FileObj.RawWrite(Buf)  ; Doesn't flush the buffer
; FileObj.Seek(0)  ; Flushes the buffer
FileObj.Close()  ; Flushes the buffer

; Length := FileObj.Length  ; Doesn't flush the buffer
; FileObj.Length := FileObj.Length  ; Flushes the buffer
; Pos := FileObj.Pos  ; Doesn't flush the buffer
; AtEOF := FileObj.AtEOF  ; Doesn't flush the buffer
; Encoding := FileObj.Encoding  ; Doesn't flush the buffer
; Handle := FileObj.Handle  ; Flushes the buffer

if Filename != '*'
   Run Filename
MsgBox
if Filename != '*'
   Run Filename
```
[Ref 1](https://github.com/AutoHotkey/AutoHotkey/blob/47d2924b932d844cf68c6ec04d55a3201050cd10/source/TextIO.cpp#L747)

I verified these results by looking at the following source code files:

- TextIO.h - Search for the FlushWriteBuffer() function.
- TextIO.cpp - Search for the PrepareToRead() function.

Clearly, there are a number of ways to manually flush the buffer. Thus, making the change I suggested in this PR could be misleading. While it probably doesn't make sense to document all of them, it would be useful to document _at least one_ in case a user has such a need. [Example <i>#</i>3](https://www.autohotkey.com/docs/v2/lib/FileOpen.htm#ExStreams) uses `stdout.Read(0)` so it may make sense to list that one in the [Remarks](https://www.autohotkey.com/docs/v2/lib/FileOpen.htm#Remarks) section of that page. I will create a separate pull request for that. For now, I will close this one.

Thank you.

PR #702, opened 2024-08-13
Update Gui.htm

iPhilip:
I am not sure if you saw [my latest comment](https://github.com/AutoHotkey/AutoHotkeyDocs/pull/698#issuecomment-2239267687) about this issue. In my opinion, the current language is ambiguous.

If you disagree, feel free to comment or close this PR.


Ragnar-F:
> In retrospect, I am concerned that "created by the same script" may lead to confusion. For example, take the above script where a WordPad window is created using `Run 'Wordpad.exe', , , &PID`. A user may assume that, because the WordPad window was created in the same script, it should work. In fact, it won't.
> 
> I would encourage you to reconsider and use "as long as the owner has the same Process ID". A link to a part of the documentation describing what a Process ID is might help those who don't know what it is.

Then it would be an incorrect assumption. The window is created by WordPad, not the script. The script only runs WordPad.

I will combine both phrases to reduce confusion.

iPhilip:
Thank you

PR #704, opened 2024-08-16
Update Gui.htm

iPhilip:
Updating the v1 documentation to reflect [PR #702](https://github.com/AutoHotkey/AutoHotkeyDocs/pull/702).


Ragnar-F:
The v1 docs will only be changed if it is really important, as the development/maintenance of v1 has now been discontinued entirely.

PR #706, opened 2024-08-26
Update CoordMode.htm

RaptorX:
Clarify what the default value is when CoordMode is never used in the documentation, because it could be missed if the person reading doesnt get to the Remarks area.

### DIFF
--- a/docs/lib/CoordMode.htm
+++ b/docs/lib/CoordMode.htm
@@ -12,7 +12,7 @@
 
 <h1>CoordMode</h1>
 
-<p>Sets coordinate mode for various built-in functions to be relative to either the active window or the screen.</p>
+<p>Sets coordinate mode for various built-in functions to be relative to either the active window or the screen. <em>Client</em> is the default CoordMode if not changed in your script.</p>
 
 <pre class="Syntax"><span class="func">CoordMode</span> TargetType <span class="optional">, RelativeTo</span></pre>
 <h2 id="Parameters">Parameters</h2>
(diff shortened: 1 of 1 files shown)


Ragnar-F:
For setting functions, the default setting is always specified under Remarks. This is also true for CoordMode:

> If this function is not used, all built-in functions except those documented otherwise (e.g. [WinMove](https://www.autohotkey.com/docs/v2/lib/WinMove.htm) and [InputBox](https://www.autohotkey.com/docs/v2/lib/InputBox.htm)) use coordinates that are relative to the active window's client area.

`If this parameter is omitted` is redundant compared to `If omitted` because it is in the Parameters section. The formatting shows exactly which parameter it refers to. It also "breaks" the general writing style of the documentation.

> ... because it could be missed if the person reading doesnt get to the Remarks area.

So you're telling me that there are people who don't manage to read the (in this case short) document in full to get the information they want? Something like RTFM would be my answer (maybe a bit more polite) 😉

RaptorX:
> > ... because it could be missed if the person reading doesnt get to the Remarks area.
> 
> So you're telling me that there are people who don't manage to read the (in this case short) document in full to get the information they want? Something like RTFM would be my answer (maybe a bit more polite) 😉

Haha, I meant it because I'm a person who reads the manual a LOT. I even read the changes from v1 to v2 even before installing v2 the first time. 

Thing is, I somehow managed to miss that information TWICE when I was trying to clarify something to a forum poster (i usually go back to the manual to confirm before affirming certain things.)

Now that you clarified it, I would amend my whole position: 
I think all functions/commands should state what the default value is right in the top description line instead of just in the remarks. Just  like the `SendMode` command which has a description like this:

```
Makes Send synonymous with SendEvent or SendPlay **rather than the default (SendInput).**
Also makes Click and MouseMove/Click/Drag use the specified method.
```

I think it is perfect that even if I just read the first line, i know what the default is for it.

That's a lot of work but I'm willing to do it. I'm a firm believer that important stuff should be mentioned at the top and prominently. Not obscured in lower sections of the documentation.

Ragnar-F:
For readers who are still beginners, this would be disadvantageous if they have to deal with secondary information while reading the description. The description should be written as simple as possible.

Since you now know that this information can always be found in the Remarks section, such changes are not necessary. This information is clearly visible, not obscured as you wrote. The most I could offer is to change the sentence to the following:

> If CoordMode is not used, the default mode is _Client_; that is, all built-in functions except those documented otherwise (e.g. [WinMove](https://www.autohotkey.com/docs/v2/lib/WinMove.htm) and [InputBox](https://www.autohotkey.com/docs/v2/lib/InputBox.htm)) use coordinates that are relative to the active window's client area.

RaptorX:
That sounds good. :)

PR #711, opened 2024-09-23
Update Hotkeys.htm

iPhilip:
Added specific requirements for the [Named Function Hotkeys](https://www.autohotkey.com/docs/v2/Hotkeys.htm#Function) to function in SciTE and Notepad++.


Ragnar-F:
This information is too specific and irrelevant to the subject. Also, the requirements may change in the future, increasing the maintenance effort. I will add '(may require title bar setting changes)'. This is sufficient.

PR #716, opened 2024-11-12
Update LoadPicture.htm

iPhilip:
1. Added a reference to the use of negative icon index values (resource IDs) to the Options section. This is the same language used in the TraySetIcon documentation page.
2. Added a sentence to the Return Value section to indicate that zero is returned if there are any errors.


Ragnar-F:
Thanks. Please change it so that the "If negative" sentence follows the "For example" sentence, as the "For example" sentence refers to the first sentence of the paragraph.

iPhilip:
Done. Take a look.

PR #722, opened 2024-11-29
Update DetectHiddenWindows.htm

RaptorX:
This exception can be found only in the WinTitle Help section on [ahk_ids](https://www.autohotkey.com/docs/v2/misc/WinTitle.htm#ahk_id).


Ragnar-F:
This information is already available in the Remarks section.

PR #731, opened 2025-01-15
Update CallbackCreate.htm

iPhilip:
Added a line to Example # 2 to remove the subclass from the control. Doing so prevents the function from having to process further messages [[Reference](https://learn.microsoft.com/en-us/windows/win32/winmsg/using-window-procedures#subclassing-a-window)].


iPhilip:
The edits don't allow the custom background color to be displayed when the window is refreshed. Thus, I am closing this pull request.

PR #739, opened 2025-05-05
Update WinGetList Example

RaptorX:
The original example script loops through the windows and shows a msgbox that the user have to click per window.

This is a refactor that uses a listview instead of forcing clicking on each window to continue. 

This change makes the result of the script a little bit clearer for a new user.

### DIFF
--- a/docs/lib/WinGetList.htm
+++ b/docs/lib/WinGetList.htm
@@ -43,25 +43,22 @@
 <h2 id="Examples">Examples</h2>
 <div class="ex" id="ExBasic">
 <p><a class="ex_number" href="#ExBasic"></a> Visits all windows on the entire system and displays info about each of them.</p>
-<pre>ids := WinGetList(,, "Program Manager")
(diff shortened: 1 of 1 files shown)


Ragnar-F:
Thanks for your efforts, but your script is just an alternative of the current example. To claim that it's clearer than the current one is just speculation. Some people might tell you that the current example is easier to understand. For instance, you can visually see how the current example goes through the windows – which yours doesn't.

RaptorX:
> Thanks for your efforts, but your script is just an alternative of the current example. To claim that it's clearer than the current one is just speculation. Some people might tell you that the current example is easier to understand. For instance, you can visually see how the current example goes through the windows – which yours doesn't.

I understand. No worries.

Just to clarify, i did mention that "__the **result** of the script__" was what was clearer meaning that in the end you could see a list of all the windows that were actually captured. Not that the script itself is clearer because I agree with you that that is a very subjective thing :)

PR #765, opened 2025-12-27
Update Functions.htm

iPhilip:
Modified the example to be more consistent with its intended demonstration purpose: non-static local variables of the outer function are ignored.


Lexikos:
This is not useful, and misses part of the original intent.

The assignment already proves that non-local variables of the outer function are ignored.

Displaying x immediately after assigning it proves nothing. The inner MsgBox would display whatever value was just assigned, regardless of whether it was assigned to the outer x or the inner x.

The purpose of the inner MsgBox is to show that the function is `Func` and not `Closure`.

PR #767, opened 2026-01-14
Update GuiControls.htm

iPhilip:
Corrected language and expanded example.

The original language:

>The colors of the day numbers inside the drop-down calendar obey that set by Gui.SetFont or the c (Color) option.

is not correct. The following script demonstrates that:

```
MyGui := Gui()
MyGui.SetFont("cRed")
MyGui.Add("DateTime", "cRed")
MyGui.Show()
```

Hopefully, the expanded example will make it easier for people to add color to their drop-down calendars.


Ragnar-F:
Thanks! I'll make a few adjustments and add something similar for MonthCal. As a side note, the quote is correct – as long as you use WinXP 😉

iPhilip:
> As a side note, the quote is correct – as long as you use WinXP 😉

That's good to know! I believe only Windows 7 and later are supported in AutoHotkey v2.

PR #768, opened 2026-01-21
Update ControlShowDropDown.htm

iPhilip:
It also works with DropDownList controls.

### DIFF
--- a/docs/lib/ControlShowDropDown.htm
+++ b/docs/lib/ControlShowDropDown.htm
@@ -14,7 +14,7 @@
 
 <h1>ControlShowDropDown</h1>
 
-<p>Shows the drop-down list of a ComboBox control.</p>
+<p>Shows the drop-down list of a ComboBox or DropDownList control.</p>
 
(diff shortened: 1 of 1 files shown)


Ragnar-F:
Note that a DDL is a ComboBox with the CBS_DROPDOWNLIST style. The ClassNN of both controls starts with "ComboBox". All Control functions are intended for use with external controls rather than AHK's GUI controls, since there is a "built-in" GUI equivalent for almost every Control function. I would leave the text as it is, but the equivalents could be mentioned in some places. ControlShowDropDown and ControlHideDropDown could also be mentioned in GuiControls.htm (for both DDL and ComboBox). I'll think about it.

iPhilip:
> Note that a DDL is a ComboBox with the CBS_DROPDOWNLIST style. The ClassNN of both controls starts with "ComboBox".

That's true but less experienced users may not be aware of that.

> ... since there is a "built-in" GUI equivalent for almost every Control function.

I am not aware of a "built-in" GUI equivalent for ControlShowDropDown.

> I'll think about it.

Thank you.

PR #769, opened 2026-02-07
Update Map.htm

iPhilip:
Corrected the verb tense to be consistent with the description in other methods.

### DIFF
--- a/docs/lib/Map.htm
+++ b/docs/lib/Map.htm
@@ -103,10 +103,10 @@ MapObj := Map.<span class="func">Call</span>(Key1, Value1, Key2, Value2, ...)</p
 <pre class="Syntax">Value := MapObj.<span class="func">Get</span>(Key <span class="optional">, Default</span>)</pre>
 <p>This method does the following:</p>
 <ul>
-  <li>Return the value associated with <em>Key</em>, if found.</li>
-  <li>Return the value of the <em>Default</em> parameter, if specified.</li>
-  <li>Return the value of <code>MapObj.Default</code>, if defined.</li>
-  <li>Throw an <a href="Error.htm#UnsetError">UnsetItemError</a>.</li>
+  <li>Returns the value associated with <em>Key</em>, if found.</li>
+  <li>Returns the value of the <em>Default</em> parameter, if specified.</li>
+  <li>Returns the value of <code>MapObj.Default</code>, if defined.</li>
+  <li>Throws an <a href="Error.htm#UnsetError">UnsetItemError</a>.</li>
 </ul>
 <p>When <em>Default</em> is omitted, this is equivalent to <code>MapObj[Key]</code>, except that <a href="#__Item">__Item</a> is not called.</p>
 </div>


Lexikos:
FYI, the original verb tense was intentional and correct. It used the *imperative* tense because the method is following a series of instructions, as introduced by:
> This method does the following

*Present* tense feels wrong to me, probably because the steps are evaluated in sequence. That implies they are at different times and can't all be in the present.

It was already consistent with the description in *Get (Array)*; as of now it is not.

Ragnar-F:
Okay, I see. In that case, I'll revert it.

@Lexikos, if the sequence happens exactly in this order, what do you think about using `<ol>` (ordered/numbered list) instead of `<ul>` (unordered list) to make this clear?

iPhilip:
> FYI, the original verb tense was intentional and correct. It used the _imperative_ tense because the method is following a series of instructions, as introduced by:
> 
> > This method does the following
> 
> _Present_ tense feels wrong to me, probably because the steps are evaluated in sequence. That implies they are at different times and can't all be in the present.
> 
> It was already consistent with the description in _Get (Array)_; as of now it is not.

In my opinion, the current sentence structure does not support the imperative tense. As written, the list describes what the method does, i.e. the method returns .... (not the method return ...). Perhaps, there is another way to structure the sentence. I cannot think of one at this time.

Lexikos:
@Ragnar-F I don't think it makes much difference, but I do not object.

Would this be clearer?
> This method carries out the following steps:

@iPhilip 
I think "Returns x." is technically not a complete sentence, as it lacks a subject. By contrast, an imperative sentence usually omits the subject. I think it is ironic to argue that the sentence structure doesn't support imperative tense, when imperative tense supports the sentence structure and present tense does not. (I prefer "Returns x." over "This method returns x.", which seems stiff and unnecessarily verbose, taking longer to get to the useful part. This is a reference for AutoHotkey users, not schoolwork or a formal document.)

If you look at it as an abbreviation, the missing part can be either "This method" or "This method does". Some of these are unnatural.
  - (It does) return a value.
  - (It does) returns a value.
  - (This method) return a value.
  - (This method) returns a value.

I had written some new documentation for Array Get (intentionally not yet pushed), and then went to copy-paste it into Map and found that it looks out of place. I did not want to change the new sentence to match and leave the two pages inconsistent. I did not want to change the Array page because I (naturally) think it sounds better the way I originally wrote it. I did not want to merely revert the Map page... well, now I'm not sure why I didn't do that.

I only wrote this much because I've been forced to form an opinion, not because it is important to me.

iPhilip:
> @Ragnar-F I don't think it makes much difference, but I do not object.
> 
> Would this be clearer?
> 
> > This method carries out the following steps:
> 

> @iPhilip I think "Returns x." is technically not a complete sentence, as it lacks a subject.

I agree. As you stated below it assumes an "It" at the beginning of each sentence. As such, it's an abbreviation.

> By contrast, an imperative sentence usually omits the subject.

Out of curiosity, I asked ChatGPT to restate the paragraph using the imperative tense. This is what it produced:

Use this method to:

- Return the value associated with Key if found.
- Return the value of the Default parameter if specified.
- Return the value of MapObj.Default if defined.
- Throw an UnsetItemError otherwise.

It also stated that

"The imperative tense (more precisely, imperative mood) is used to give commands, instructions, requests, or directions. It tells someone what to do."

My impression was that the page was describing what the method does (which I prefer), not telling someone what to do. Perhaps, my impression was wrong.

I still prefer the "This method does the following: - Returns ..." form.

> I think it is ironic to argue that the sentence structure doesn't support imperative tense, when imperative tense supports the sentence structure and present tense does not. (I prefer "Returns x." over "This method returns x.", which seems stiff and unnecessarily verbose, taking longer to get to the useful part. This is a reference for AutoHotkey users, not schoolwork or a formal document.)

I agree. I also don't like verbose language. My argument was not well thought-out. The sentence structure has nothing to do with it.

> 
> If you look at it as an abbreviation, the missing part can be either "This method" or "This method does". Some of these are unnatural.
> 
>     * (It does) return a value.
> 
>     * (It does) returns a value.
> 
>     * (This method) return a value.
> 
>     * (This method) returns a value.
> 
> 
> I had written some new documentation for Array Get (intentionally not yet pushed), and then went to copy-paste it into Map and found that it looks out of place. I did not want to change the new sentence to match and leave the two pages inconsistent. I did not want to change the Array page because I (naturally) think it sounds better the way I originally wrote it. I did not want to merely revert the Map page... well, now I'm not sure why I didn't do that.
> 
> I only wrote this much because I've been forced to form an opinion, not because it is important to me.

I know you time is valuable and I appreciate everything you have done and are doing for this community. It was not my intention to force you to form an opinion. I was just pointing out something that didn't sound right to me. My hope was (and continues to be) that my suggestion would be helpful to others as well.

PR #780, opened 2026-08-17
Clarifying description of the Capacity property in the Array class

iPhilip:
Clarifies that the `Capacity` property in the Array class affects the current *allocated memory* capacity of the array. As is, the documentation does not clearly distinguish the difference between the `Length` and the `Capacity` properties. Some users are confusing the two. For example, see [this post](https://www.autohotkey.com/boards/viewtopic.php?f=82&t=141026).

The following code snippet distinguishes the two:

```
arr := [10, 20, 30]
MsgBox arr.Length    ; 3 — three elements
MsgBox arr.Capacity  ; 3 or more — depends on internal allocation

arr.Capacity := 100  ; pre-allocate for 100 elements (no reallocation needed as you push)
arr.Push(40)
MsgBox arr.Length    ; 4
MsgBox arr.Capacity  ; still 100

### DIFF
--- a/docs/lib/Array.htm
+++ b/docs/lib/Array.htm
@@ -54,7 +54,7 @@ MsgBox users[1]</pre>
   <li><a href="#Properties">Properties</a>:
     <ul>
       <li><a href="#Length">Length</a>: Gets or sets the length of an array.</li>
-      <li><a href="#Capacity">Capacity</a>: Gets or sets the current capacity of an array.</li>
+      <li><a href="#Capacity">Capacity</a>: Gets or sets the current allocated memory capacity of an array.</li>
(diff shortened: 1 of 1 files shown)


Ragnar-F:
Hmm, the distinction between Length and Capacity is actually pretty clear in the docs. Just read the descriptions carefully. However, I don't think the short description for Length is very well chosen. Something like "Gets or sets the number of items in an array." would be better, similar to the wording used on the [Objects](https://www.autohotkey.com/docs/v2/Objects.htm#Usage_Simple_Arrays) page.

iPhilip:
I agree. Thank you for suggesting that.

